PR #12, opened 2023-06-21
Update Registry.sol

### LINKED ISSUE #14: Consolidate Registry.sol
per #11 

- merge concepts and comments _from_ IRegistry.sol _into_ Registry.sol
- Remove IRegistry.sol

codenamejason:
### Description

Build out Registry.sol

Resolves #14

### DIFF
--- a/contracts/core/Registry.sol
+++ b/contracts/core/Registry.sol
@@ -9,13 +9,16 @@ contract Registry is AccessControl {
     /// @notice Custom errors
     error NO_ACCESS_TO_ROLE();
     error NONCE_NOT_AVAILABLE();
+    error NO_PENDING_OWNER();
+    error NOT_PENDING_OWNER();
 
     /// @notice Struct to hold details of an identity
     struct Identity {
         uint256 nonce;
-        address owner;
         string name;
         Metadata metadata;
+        address owner;
+        address pendingOwner;
         address anchor;
     }
 
@@ -39,6 +42,7 @@ contract Registry is AccessControl {
     event IdentityNameUpdated(bytes32 indexed identityId, string name, address anchor);
     event IdentityMetadataUpdated(bytes32 indexed identityId, Metadata metadata);
     event IdentityOwnerUpdated(bytes32 indexed identityId, address owner);
+    event IdentityPendingOwnerUpdated(bytes32 indexed identityId, address pendingOwner);
 
     /// ====================================
     /// =========== Modifier ===============
@@ -84,15 +88,16 @@ contract Registry is AccessControl {
     ) external returns (bytes32) {
         bytes32 identityId = _generateIdentityId(_nonce);
 
-        if (identitiesById[identityId].owner != address(0)) {
+        if (identitiesById[identityId].anchor != address(0)) {
             revert NONCE_NOT_AVAILABLE();
         }
 
         Identity memory identity = Identity({
             nonce: _nonce,
-            owner: _owner,
             name: _name,
             metadata: _metadata,
+            owner: _owner,
+            pendingOwner: address(0),
             anchor: _generateAnchor(identityId, _name)
         });
 
@@ -176,13 +181,35 @@ contract Registry is AccessControl {
         return hasRole(_identityId, _member);
     }
 
-    /// @notice Transfers the ownership of the identity to a new owner
+    /// @notice Updates the pending owner of the identity
     /// @param _identityId The identityId of the identity
-    /// @param _owner New Owner
-    /// @dev Only owner can transfer ownership.
-    function changeIdentityOwner(bytes32 _identityId, address _owner) external isIdentityOwner(_identityId) {
-        identitiesById[_identityId].owner = _owner;
-        emit IdentityOwnerUpdated(_identityId, _owner);
+    /// @param _pendingOwner New pending owner
+    function updateIdentityPendingOwner(bytes32 _identityId, address _pendingOwner)
+        external
+        isIdentityOwner(_identityId)
+    {
+        identitiesById[_identityId].pendingOwner = _pendingOwner;
+
+        emit IdentityPendingOwnerUpdated(_identityId, _pendingOwner);
+    }
+
+    /// @notice Transfers the ownership of the identity to the pending owner
+    /// @param _identityId The identityId of the identity
+    /// @dev Only pending owner can claim ownership.
+    function changeIdentityOwner(bytes32 _identityId) external {
+        Identity storage identity = identitiesById[_identityId];
+
+        if (identity.pendingOwner == address(0)) {
+            revert NO_PENDING_OWNER();
+        }
+        if (msg.sender != identity.pendingOwner) {
+            revert NOT_PENDING_OWNER();
+        }
+
+        identity.owner = identity.pendingOwner;
+        identity.pendingOwner = address(0);
+
+        emit IdentityOwnerUpdated(_identityId, identity.owner);
     }
 
     /// @notice Adds members to the identity
--- a/contracts/core/Registry.sol
+++ b/contracts/core/Registry.sol
@@ -19,14 +19,14 @@ contract Registry is AccessControl {
     }
 
     /// @notice Struct to hold details of an identity
-    struct IdentityDetails {
+    struct Identity {
         string name;
         Metadata metadata;
         address anchor;
     }
 
-    /// @notice identityId -> IdentityDetails
-    mapping(uint => IdentityDetails) public identities;
+    /// @notice identityId -> Identity
+    mapping(uint => Identity) public identities;
 
     // Events
     event IdentityCreated(
@@ -44,23 +44,23 @@ contract Registry is AccessControl {
 
     /// @notice Creates a new identity
     /// @dev This will also set the attestation address generated from msg.sender and name
-    /// @param name The name of the identity
-    /// @param metadata The metadata of the identity
+    /// @param _name The name of the identity
+    /// @param _metadata The metadata of the identity
     /// @param _owners The owners of the identity
     /// @param _members The members of the identity
     function createIdentity(
-        string memory name,
-        Metadata memory metadata,
+        string memory _name,
+        Metadata memory _metadata,
         address[] memory _owners,
         address[] memory _members
     ) external returns (uint256) {
-        IdentityDetails memory identityDetails = IdentityDetails(
-            name,
-            metadata,
-            _generateAnchor(name, msg.sender)
+        Identity memory identity = Identity(
+            _name,
+            _metadata,
+            _generateAnchor(_name, msg.sender)
         );
 
-        identities[identityId] = identityDetails;
+        identities[identityId] = identity;
 
         // generate roles
         bytes32 ownerRole = _generateRole(identityId, RoleType.OWNER);
@@ -75,12 +75,12 @@ contract Registry is AccessControl {
             _grantRole(memberRole, _members[i]);
         }
 
-        // NOTE: should we use the identityDetails we created above or the data passed in?
+        // NOTE: should we use the identity we created above or the data passed in?
         emit IdentityCreated(
             identityId,
-            name,
-            metadata,
-            identityDetails.anchor
+            _name,
+            _metadata,
+            identity.anchor
         );
 
         identityId++;
@@ -222,7 +222,7 @@ contract Registry is AccessControl {
         _grantRole(role, account);
 
         if (_roleType == RoleType.OWNER) {
-            IdentityDetails memory identity = identities[_identityId];
+            Identity memory identity = identities[_identityId];
             // TODO: HOW to update attestation address cause name and identityId is never updated?
             // NOTE: see comment: https://github.com/allo-protocol/allo-v2/pull/12#discussion_r1238843703
         }
@@ -242,7 +242,7 @@ contract Registry is AccessControl {
         bytes32 role = _generateRole(_identityId, _roleType);
         _revokeRole(role, account);
         if (_roleType == RoleType.OWNER) {
-            IdentityDetails memory identity = identities[_identityId];
+            Identity memory identity = identities[_identityId];
             // TODO: How to update attestation address cause name and identityId is never updated?
             // NOTE: see comment: https://github.com/allo-protocol/allo-v2/pull/12#discussion_r1238843703
         }
--- a/contracts/core/Registry.sol
+++ b/contracts/core/Registry.sol
@@ -53,16 +53,15 @@ contract Registry is AccessControl {
     /// @param _index Index of the identity
     /// @param _name The name of the identity
     /// @param _metadata The metadata of the identity
-    /// @param _owners The owners of the identity
+    /// @param _owner The owner of the identity
     /// @param _members The members of the identity
     function createIdentity(
         uint _index,
         string memory _name,
         Metadata memory _metadata,
-        address[] memory _owners,
+        address _owner, // Note: this is a single owner
         address[] memory _members
     ) external returns (bytes32) {
-
         bytes32 identityId = _generateIdentityId(_index);
 
         if (identitiesById[identityId].id == bytes32(0)) {
@@ -85,17 +84,12 @@ contract Registry is AccessControl {
         bytes32 memberRole = _generateRole(identityId, RoleType.MEMBER);
 
         // assign roles
-        uint256 ownerLength = _owners.length;
-        for (uint i = 0; i < ownerLength;) {
-            address _owner = _owners[i];
-            _grantRole(ownerRole, _owner);
-            _grantRole(memberRole, _owner);
-            unchecked {
-                i++;
-            }
-        }
+        address owner = _owner;
+        _grantRole(ownerRole, owner);
+        _grantRole(memberRole, owner);
+
         uint256 memberLength = _members.length;
-        for (uint i = 0; i < memberLength;) {
+        for (uint i = 0; i < memberLength; ) {
             _grantRole(memberRole, _members[i]);
             unchecked {
                 i++;
@@ -116,7 +110,7 @@ contract Registry is AccessControl {
     /// @notice Updates the name of the identity and generates new anchor
     /// @param _identityId The identityId of the identity
     /// @param _name The new name of the identity
-    /// @dev Only owner can update the name. _identityId is reused. 
+    /// @dev Only owner can update the name. _identityId is reused.
     function updateIdentityName(
         bytes32 _identityId,
         string memory _name
@@ -127,11 +121,8 @@ contract Registry is AccessControl {
             revert NO_ACCESS_TO_ROLE();
         }
 
-        address anchor = _generateAnchor(
-            _identityId,
-            _name
-        );
-        
+        address anchor = _generateAnchor(_identityId, _name);
+
         Identity storage identity = identitiesById[_identityId];
         identity.name = _name;
         identity.anchor = anchor;
@@ -174,13 +165,13 @@ contract Registry is AccessControl {
 
     /// @notice Returns if the given address is an member of the identity
     /// @param _identityId The identityId of the identity
-    /// @param _owner The address to check
+    /// @param _member The address to check
     function isMemberOfIdentity(
         bytes32 _identityId,
-        address _owner
+        address _member
     ) external view returns (bool) {
         bytes32 memberRole = _generateRole(_identityId, RoleType.MEMBER);
-        return hasRole(memberRole, _owner);
+        return hasRole(memberRole, _member);
     }
 
     /// @notice Generates the anchor for the given identityId and name
@@ -190,7 +181,9 @@ contract Registry is AccessControl {
         bytes32 _identityId,
         string memory _name
     ) internal pure returns (address) {
-        bytes32 attestationHash = keccak256(abi.encodePacked(_identityId, _name));
+        bytes32 attestationHash = keccak256(
+            abi.encodePacked(_identityId, _name)
+        );
 
         return address(uint160(uint256(attestationHash)));
     }
@@ -198,12 +191,7 @@ contract Registry is AccessControl {
     /// @notice Generates the identityId based on msg.sender
     /// @param _index Index of the identity
     function _generateIdentityId(uint _index) internal view returns (bytes32) {
-        return keccak256(
-            abi.encodePacked(
-                msg.sender,
-                _index
-            )
-        );
+        return keccak256(abi.encodePacked(msg.sender, _index));
     }
 
     /// @notice Generates the OZ role for an given identity
@@ -216,8 +204,57 @@ contract Registry is AccessControl {
         roleHash = keccak256(abi.encodePacked(_identityId, _roleType));
     }
 
-    /// TODO: Role management
-    /// Who can add / remove owner for an identity
-    /// Who can add / remove members for an identity
-    /// Multiple owners vs single owner
+    // Note: Role management - Spoke with Nate, we are going to go with single owner for now
+    // ! Who can add / remove owner for an identity? No one. It's a single owner.
+    // ? Transfer ownership will not be supported for now.
+    // - Owner can add / remove members
+    // - Single owner
+
+    /// @notice Adds members to the identity
+    /// @param _identityId The identityId of the identity
+    /// @param _members The members to add
+    function addMembers(
+        bytes32 _identityId,
+        address[] memory _members
+    ) external {
+        bytes32 ownerRole = _generateRole(_identityId, RoleType.OWNER);
+
+        if (!hasRole(ownerRole, msg.sender)) {
+            revert NO_ACCESS_TO_ROLE();
+        }
+
+        bytes32 memberRole = _generateRole(_identityId, RoleType.MEMBER);
+
+        uint256 memberLength = _members.length;
+        for (uint i = 0; i < memberLength; ) {
+            _grantRole(memberRole, _members[i]);
+            unchecked {
+                i++;
+            }
+        }
+    }
+
+    /// @notice Removes members from the identity
+    /// @param _identityId The identityId of the identity
+    /// @param _members The members to remove
+    function removeMembers(
+        bytes32 _identityId,
+        address[] memory _members
+    ) external {
+        bytes32 ownerRole = _generateRole(_identityId, RoleType.OWNER);
+
+        if (!hasRole(ownerRole, msg.sender)) {
+            revert NO_ACCESS_TO_ROLE();
+        }
+
+        bytes32 memberRole = _generateRole(_identityId, RoleType.MEMBER);
+
+        uint256 memberLength = _members.length;
+        for (uint i = 0; i < memberLength; ) {
+            _revokeRole(memberRole, _members[i]);
+            unchecked {
+                i++;
+            }
+        }
+    }
 }
--- a/contracts/core/Registry.sol
+++ b/contracts/core/Registry.sol
@@ -20,7 +20,7 @@ contract Registry is AccessControl {
     /// @notice Struct to hold details of an identity
     struct Identity {
         bytes32 id;
-        uint index;
+        uint nonce;
         string name;
         Metadata metadata;
         address anchor;
@@ -42,7 +42,7 @@ contract Registry is AccessControl {
 
     event IdentityCreated(
         bytes32 indexed identityId,
-        uint index,
+        uint nonce,
         string name,
         Metadata metadata,
         address anchor
@@ -83,19 +83,19 @@ contract Registry is AccessControl {
 
     /// @notice Creates a new identity
     /// @dev This will also set the attestation address generated from msg.sender and name
-    /// @param _index Index of the identity
+    /// @param _nonce Index of the identity
     /// @param _name The name of the identity
     /// @param _metadata The metadata of the identity
     /// @param _owner The owner of the identity
     /// @param _members The members of the identity
     function createIdentity(
-        uint _index,
+        uint _nonce,
         string memory _name,
         Metadata memory _metadata,
         address _owner,
         address[] memory _members
     ) external returns (bytes32) {
-        bytes32 identityId = _generateIdentityId(_index);
+        bytes32 identityId = _generateIdentityId(_nonce);
 
         if (identitiesById[identityId].id != bytes32(0)) {
             revert INDEX_NOT_AVAILABLE();
@@ -103,7 +103,7 @@ contract Registry is AccessControl {
 
         Identity memory identity = Identity(
             identityId,
-            _index,
+            _nonce,
             _name,
             _metadata,
             _generateAnchor(identityId, _name)
@@ -298,9 +298,9 @@ contract Registry is AccessControl {
     }
 
     /// @notice Generates the identityId based on msg.sender
-    /// @param _index Index of the identity
-    function _generateIdentityId(uint _index) internal view returns (bytes32) {
-        return keccak256(abi.encodePacked(_index, msg.sender));
+    /// @param _nonce Index of the identity
+    function _generateIdentityId(uint _nonce) internal view returns (bytes32) {
+        return keccak256(abi.encodePacked(_nonce, msg.sender));
     }
 
     /// @notice Generates the OZ role for an given identity
--- a/contracts/core/Registry.sol
+++ b/contracts/core/Registry.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.8.19;
 
-import {AccessControl} from "../../lib/openzeppelin-contracts/contracts/access/AccessControl.sol";
+import {AccessControl} from "openzeppelin-contracts/contracts/access/AccessControl.sol";
 
 import {Metadata} from "./libraries/Metadata.sol";
 
--- a/contracts/core/Registry.sol
+++ b/contracts/core/Registry.sol
@@ -59,6 +59,7 @@ contract Registry is AccessControl {
     /// ====================================
     /// ========== Constructor =============
     /// ====================================
+
     constructor(address _admin) {
         // DEFAULT_ADMIN_ROLE would be set by Allo team
         // to grant or revoke roles in emergencies.
@@ -66,9 +67,10 @@ contract Registry is AccessControl {
     }
 
     /// ====================================
-    /// ========== Modifier =============
+    /// =========== Modifier ===============
     /// ====================================
-    modifier isPoolOwner(bytes32 _identityId) {
+
+    modifier isIdentityOwner(bytes32 _identityId) {
         if (!isOwnerOfIdentity(_identityId, msg.sender)) {
             revert NO_ACCESS_TO_ROLE();
         }
@@ -155,15 +157,18 @@ contract Registry is AccessControl {
     function updateIdentityName(
         bytes32 _identityId,
         string memory _name
-    ) external isPoolOwner(_identityId) returns (address) {
+    ) external isIdentityOwner(_identityId) returns (address) {
 
         address anchor = _generateAnchor(_identityId, _name);
 
         Identity storage identity = identitiesById[_identityId];
         identity.name = _name;
-        identity.anchor = anchor;
 
-        // TODO: should we clear old anchor?
+        // clear old anchor
+        anchorToIdentityId[identity.anchor] = bytes32(0);
+
+        // set new anchor
+        identity.anchor = anchor;
         anchorToIdentityId[identity.anchor] = _identityId;
 
         emit IdentityNameUpdated(_identityId, _name, anchor);
@@ -179,7 +184,7 @@ contract Registry is AccessControl {
     function updateIdentityMetadata(
         bytes32 _identityId,
         Metadata memory _metadata
-    ) external isPoolOwner(_identityId) {
+    ) external isIdentityOwner(_identityId) {
 
         identitiesById[_identityId].metadata = _metadata;
 
@@ -237,7 +242,7 @@ contract Registry is AccessControl {
     function addMembers(
         bytes32 _identityId,
         address[] memory _members
-    ) external isPoolOwner(_identityId) {
+    ) external isIdentityOwner(_identityId) {
 
         bytes32 memberRole = _generateRole(_identityId, RoleType.MEMBER);
 
@@ -257,7 +262,7 @@ contract Registry is AccessControl {
     function removeMembers(
         bytes32 _identityId,
         address[] memory _members
-    ) external isPoolOwner(_identityId) {
+    ) external isIdentityOwner(_identityId) {
 
         bytes32 memberRole = _generateRole(_identityId, RoleType.MEMBER);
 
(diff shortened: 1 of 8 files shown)


zobront:
Yep, this is good. We'll probably want an `id => attestation address` mapping we maintain too, so that we are able to look up identities by id or attestation address, but it's fine with me if address is the main one.

codenamejason:
@zobront I only did the first item so far from this list:

- [ ] No linked lists
- [ ] Solmate Roles (with each identity having two types of roles, owner and member)
- [ ] IdentityID is just an incrementer
- [ ] attestationAddress is `address(uint160(uint(keccak256(identityId, name))))`

thelostone-mc:
@codenamejason I'd say let's create the registry.sol and take it from there to make it easier to build

contracts/core/Registry.sol
```diff
@@ -1,242 +1,110 @@
...
-    // The mapping of identities, from identityId to IdentityDetails
     mapping(uint => IdentityDetails) public identities;
 
-    // The mapping identities owners, from identityId to OwnerList
-    mapping(address => OwnerList) public identityOwners;
-
     // Events
+    event IdentityCreated(uint indexed identityId, string name, Metadata metadata, address attestationAddr);
+    event IdentityNameUpdated(uint indexed identityId, string name);
+    event IdentityMetadataUpdated(uint indexed identityId, Metadata metadata);
```
thelostone-mc, line 24: Does it make sense to have multiple events ? This was keeping in mind for easier indexing

contracts/core/Registry.sol
```diff
@@ -1,242 +1,110 @@
...
+    uint private identityId;
 
     struct IdentityDetails {
-        uint id;
         string name;
-        Metadata.MetaPtr metadata;
-        address attestationAddress;
+        Metadata metadata;
+        address attestationAddr;
+        // address coreOwner; // ASK: Should this be added?
```
thelostone-mc, line 16: Since this contract uses Roles, we can use solady to track owners but how do we track coreOwner without capturing it separately ?

contracts/core/Registry.sol
```diff
@@ -1,18 +1,33 @@
...
         string name;
-        string metadata; // @todo update to MetaPtr
+        Metadata metadata;
         address attestationAddr;
+        // address coreOwner; // ASK: Should this be added?
     }
 
-    // getter for projects mapping
-    function identities(uint _identityId) external view returns (IdentityDetails memory) {
+    mapping(uint => IdentityDetails) public identities;
```
thelostone-mc, line 19: Would it make sense to have identities as a public mapping instead and not having the identities getter function ?

contracts/core/Registry.sol
```diff
@@ -21,27 +36,75 @@ contract Registry {
...
+        Metadata memory metadata,
+        address[] memory _owners
+    ) external returns (uint256) {
+
+        IdentityDetails memory identityDetails = IdentityDetails(
+            name,
+            metadata,
+            // ASK : should identityId be used as the hash for attestationAddr?
+            // cause project can't have same attestAddr across chains
+            generateAttestationAddr(name, msg.sender)
```
thelostone-mc, line 54: @zobront Do we need the counter to generate the attestation address  ? Check comment above

contracts/core/Registry.sol
```diff
@@ -21,27 +36,75 @@ contract Registry {
...
+        );
+
+        identities[identityId++] = identityDetails;
+
+        for(uint i = 0; i < _owners.length; i++) {
+            _grantRoles(_owners[i], identityId);
```
thelostone-mc, line 60: identityId is being used as the Role here.

contracts/core/Registry.sol
```diff
@@ -21,27 +36,75 @@ contract Registry {
...
+        identities[_identityId].name = _name;
+        identities[_identityId].attestationAddr = generateAttestationAddr(_name, msg.sender);
+
+        emit IdentityNameUpdated(_identityId, _name);
     }
 
-    // updates the name of the identity
-    // note that this will also change the attestation address, since it's a hash that includes the name
-    function updateIdentityName(uint _identityId, string memory _name) external {
+    function changeCoreOwner(uint _identityId, address _newCoreOwner) external {
```
thelostone-mc, line 77: Not sure how to track core owner of a identity separately. 
If we store it IdentityDetails, then this function is needed 

@zobront

contracts/core/libraries/Metadata.sol
```diff
@@ -1,12 +1,10 @@
...
-library Metadata {
-    struct MetaPtr {
-        /// @notice Protocol ID corresponding to a specific protocol.
-        /// More info at https://github.com/allo-protocol/contracts/tree/main/docs/MetaPtrProtocol.md
-        uint256 protocol;
-        /// @notice Pointer to fetch metadata for the specified protocol
-        string pointer;
-    }
-}
+struct Metadata {
```
thelostone-mc, line 4: @codenamejason I removed the library to avoid having to write `Metadata.MetaPtr` everywhere. Lemme know if we want to change this back to an actual library

thelostone-mc (commented):
wip but left a few open questions

contracts/core/Registry.sol
```diff
@@ -21,27 +36,75 @@ contract Registry {
...
 
+        emit IdentityMetadataUpdated(_identityId, _metadata);
     }
 
-    // this will use solmate ROLES to check if the msg.sender is an owner of the project
-    // @todo figure out how to best represent project ownership and details
-    function isOwnerOfProject(uint _identityId, address _owner) external view returns (bool) {
-        
+    /// @notice Returns if the given address is an owner of the project
+    function isOwnerOfIdentity(uint _identityId, address _owner) external view returns (bool) {
```
thelostone-mc, line 98: renaming this to `isOwnerOfIdentity` for consistency.
Also do we need this function ? Can't we just expect folks to use hasAnyRole directly ?

contracts/core/Registry.sol
```diff
@@ -1,18 +1,35 @@
...
 
+import { OwnableRoles } from "../../lib/solady/src/auth/OwnableRoles.sol";
+import { Metadata } from "./libraries/Metadata.sol";
+
 // use Solady Roles for ownership of projects: https://github.com/Vectorized/solady/blob/main/src/auth/OwnableRoles.sol
-contract Registry {
+contract Registry is OwnableRoles {
+
+    // ASK: should this be renamed to identityIndex ?
+    uint private identityId;
```
thelostone-mc, line 11: @zobront does it make sense to rename this to identityIndex

contracts/core/Registry.sol
```diff
@@ -21,27 +38,77 @@ contract Registry {
...
+        }
+        
+        emit IdentityCreated(identityId, name, metadata, identityDetails.attestationAddr);
+        
+        return identityId;
+    }
+
+    /// @notice Updates the name of the identity
+    /// @dev This will also change the attestation address, since it's a hash that includes the name
+    // ASK: should only coreOwner be able to change name?
```
thelostone-mc, line 72: ^ @zobront @codenamejason

contracts/core/Registry.sol
zobront: I don't see why not. I like these three.

contracts/core/Registry.sol
zobront: Do we want coreOwner? I thought on the call yesterday we decided we were fine without the ability to look up owner, because the legible version of who they are is captured in the name and metadata, so owner would just be an address that wouldn't add much. 

If we decide it's needed, agreed it should be added to this struct.

contracts/core/Registry.sol
zobront: Yes, agreed.

contracts/core/Registry.sol
zobront: I don't think `msg.sender` should be a part of this. We want the attestation address to be consistent as long as the name doesn't change. 

I think `identityId` (immutable) and `name` is the right thing.

I don't think it's a problem to have different attestation addresses across chains, because we'll need to be attesting separately in every chain anyways.

contracts/core/Registry.sol
zobront: We want to add an extra variable because there are two types of roles, owner and member. 

Owners can change the metadata.

Members can claim to be a part of the organization when creating a pool or applying.

I would create an enum for types of roles (OWNER and MEMBER) and then would `abi.encode(identityId, roleType)` to get the role for Solady.

contracts/core/Registry.sol
zobront: We could. `hasAnyRole()` uses a bitmap for the roles and requires the caller to put together the role id so may be a bit tricky for users. I like the idea of having these helper functions that make it easy to call with just the identityId and the owner.

contracts/core/Registry.sol
zobront: Either one is fine. I prefer ID but your call.

contracts/core/Registry.sol
zobront: I think there may be more than 1 person we want to give the ability to change the name.

I'd use the OWNERS in the roles the ability to do this, but not the MEMBERS.

zobront:
I'm digging deeper into Solady's Roles and I think it may not be the right solution. They use the `uint256` as a bitmap, which I believe limits them to 255 roles that can be claimed. Obviously we'll have more than that.

We could use OpenZeppelin's access control (https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/AccessControl.sol), which is a bit clunky but does fit the use case.

We could also put a simple version together ourselves, but would want to be very careful with the logic.

contracts/core/Registry.sol
thelostone-mc: but having a different attestation address for the same owner and name would mean that identity would have to rebuild attestations across chains. I was thinking of a scenario where attestations could be ported by a 3rd part.y from one chain to another

contracts/core/Registry.sol
thelostone-mc: Makes sense 🙌🏼

contracts/core/Registry.sol
thelostone-mc: Yup agreed. Dropping this for now

contracts/core/Registry.sol
thelostone-mc: sticking with ID

contracts/core/Registry.sol
```diff
@@ -1,47 +1,141 @@
...
+            revert NO_ACCESS_TO_ROLE();
+        }
+
+        identities[_identityId].name = _name;
+        identities[_identityId].attestationAddr = _generateAttestationAddr(_identityId, _name);
+
+        emit IdentityNameUpdated(_identityId, _name);
     }
 
     // @todo override changing owners (core owner, not contributors) to make sure it updates attestation address
```
thelostone-mc, line 90: @codenamejason this needs to be implemented.

contracts/core/Registry.sol
```diff
@@ -1,47 +1,141 @@
...
-        
+    /// @notice Returns if the given address is an owner of the identity
+    function isOwnerOfIdentity(uint _identityId, address _owner) external view returns (bool) {
+        bytes32 ownerRole = _generateRole(_identityId, RoleType.OWNER);
+        return hasRole(ownerRole, _owner);
+    }
+
+
+    /// @notice Returns if the given address is an member of the identity
+    function isMemberOfIdentity(uint _identityId, address _owner) external view returns (bool) {
```
thelostone-mc, line 121: @zobront Added this as well cause it makes sense to offer isRolecheck for both roles

contracts/core/Registry.sol
```diff
@@ -1,47 +1,174 @@
...
+        revert NOT_SUPPORTED();
+    }
+
+    function grantIdentityRole(uint _identityId, RoleType _roleType, address account) external returns (uint) {
+        bytes32 role = _generateRole(_identityId, _roleType);
+        _grantRole(role, account);
+
+        if(_roleType == RoleType.OWNER) {
+            IdentityDetails memory identity = identities[_identityId];
+            // TODO: HOW to update attestation address cause name and identityId is never updated?
```
thelostone-mc, line 159: @zobront @codenamejason  @KurtMerbeth 
I think we would _generateAttestationAddr to take an extra parameter which relies on an address or something.

contracts/core/Registry.sol
0xZakk: Yea, if it's possible, it would be great to give people the same address on different chains. I think we would have to use `msg.sender` in order to do that?

contracts/core/Registry.sol
```diff
@@ -1,47 +1,174 @@
...
+        OWNER,
+        MEMBER
+    }
+
+    /// @notice Struct to hold details of an identity
     struct IdentityDetails {
         string name;
-        string metadata; // @todo update to MetaPtr
+        Metadata metadata;
         address attestationAddr;
```
0xZakk, line 26: Maybe we should call this `identity` instead of `attestationAddr`?

Or really just something that doesn't have `attestation` or `addr` in the name because:
1. The type is `address`, so we don't need to have the type in the variable name too
2. Attestations are just on of the things these addresses can be used for - SBTs is another

Thinking through the naming:

Conceptually, the "identity" is the combination of name, metadata, and address. So we'll want to do something like:

```solidity
identity.name;
identity.metadata;
identity.<something>;
```

Some ideas:
- location / locator
- resource
- wallet
- identifier (though not a fan of `identity.identifier`)
- recipient

contracts/core/Registry.sol
```diff
@@ -1,47 +1,174 @@
...
+        return identityId;
     }
 
-    // updates the name of the identity
-    // note that this will also change the attestation address, since it's a hash that includes the name
+    /// @notice Updates the name of the identity. 
+    /// @param _identityId The identityId of the identity
+    /// @param _name The new name of the identity
+    /// @dev Only owner can update the name. Also updated the attestation address
     function updateIdentityName(uint _identityId, string memory _name) external {
```
0xZakk, line 77: Do we want to emit and/or return the new address?

contracts/core/Registry.sol
```diff
@@ -1,47 +1,174 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.0;
 
-// use Solady Roles for ownership of projects: https://github.com/Vectorized/solady/blob/main/src/auth/OwnableRoles.sol
-contract Registry {
+import { AccessControl } from "../../lib/openzeppelin-contracts/contracts/access/AccessControl.sol";
```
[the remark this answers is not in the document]
nfrgosselin: Leaning towards OZ, but it's still an open discussion. You can see @zobront's initial thoughts here: https://github.com/allo-protocol/allo-v2/pull/12#issuecomment-1602732439

contracts/core/Registry.sol
codenamejason: ### Overview
Since the `identityId` changes on each new identity created the address would be different on each chain, we cannot guarantee what the `identityId` would be with the current code (see below):

```javascript
function _generateAttestationAddr(uint _identityId, string memory _name) internal pure returns (address) {
    bytes32 attestationHash = keccak256(
        abi.encodePacked(_identityId, _name)
    );

    return address(uint160(uint256(attestationHash)));
 }
```

### Suggested Changes

Adding in the `sender` and removing the `identityId` would allow us to make sure the address is the same on each chain. I don't think we would ever have a collision where the `_name` and `_sender` are the same within a single `Identity` 🤔 

```javascript
function _generateAttestationAddr(
    string memory _name,
    address _sender
) internal pure returns (address) {
    bytes32 attestationHash = keccak256(abi.encodePacked(_name, _sender));

    return address(uint160(uint256(attestationHash)));
}
```

Let me know what you all think 🙌 There may be something I am missing 🤔

cc: @zobront @0xZakk @thelostone-mc @KurtMerbeth

contracts/core/Registry.sol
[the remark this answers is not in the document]
codenamejason: We have updated to use OZ currently for @zobront had suggested Solady would not support our needs in the future; the uint256 they use can only hold 255 roles and we need much more than that. i.e. for one `Identity` creation we have two (2) roles being generated and one more for each owner and member you add.

contracts/core/Registry.sol
codenamejason: I agree with @0xZakk on this, I am leaning toward `ownerIdentifier` though 🤩

```javascript
struct IdentityDetails {
    string name;
    Metadata metadata;
    address ownerIdentifier;
}
```

WDYT?

contracts/core/Registry.sol
codenamejason: Good question, I don't see any harm in return data as well. We could add a return like so:

```javascript
function updateIdentityName(
    uint _identityId,
    string memory _name
) external returns (address) {
    bytes32 ownerRole = _generateRole(identityId, RoleType.OWNER);

    if (!hasRole(ownerRole, msg.sender)) {
        revert NO_ACCESS_TO_ROLE();
    }

    identities[_identityId].name = _name;
    identities[_identityId].ownerIdentifier = _generateownerIdentifier(
        _name,
        msg.sender
    );

    address ownerIdentifier = identities[_identityId].ownerIdentifier;

    emit IdentityNameUpdated(_identityId, _name, ownerIdentifier);

    return ownerIdentifier;
}
```

We could also update the emit:
```javascript
event IdentityNameUpdated(uint indexed identityId, string name, address newOwnerIdentifier);
```

contracts/core/Registry.sol
nfrgosselin: The core concept makes sense to me. I like the idea of having the name convey that it's central to identity.

Riffing on that:
- location / locator (from your original list)
- anchor
- nucleus
- locus
- kernel
- hub

contracts/core/Registry.sol
nfrgosselin: Can someone help me understand how role flexibility works here? Is it just these two roles (owner/member) or will identity owners have the flexibility to create additional roles?

codenamejason: So far we have only defined these 2 roles, but we could look into a way for the `identityOwner` to create custom roles. I think that would be more complex and needs to be thought through a bit more.

We currently use an `enum` to pass when we create the role with the `msg.sender`. The function is also just used internally to this contract and not publicly callable.

```javascript
enum RoleType {
    OWNER,
    MEMBER
}

function _generateRole(
    uint _identityId,
    RoleType _roleType
) internal pure returns (bytes32 roleHash) {
    roleHash = keccak256(abi.encodePacked(_identityId, _roleType));
}
```

We can think through how we want to make this extendable.

contracts/core/Registry.sol
codenamejason: anchor is sticking for me ⚓️

codenamejason:
I pushed my latest changes to reflect some of the conversations and also give a visual on how some of the changes (also suggestions) look in the contract.

contracts/core/Registry.sol
codenamejason: https://github.com/allo-protocol/allo-v2/pull/12#discussion_r1238885697

contracts/core/Registry.sol
```diff
@@ -1,47 +1,251 @@
...
+    /// @notice Creates a new identity
+    /// @dev This will also set the attestation address generated from msg.sender and name
+    /// @param name The name of the identity
+    /// @param metadata The metadata of the identity
+    /// @param _owners The owners of the identity
+    /// @param _members The members of the identity
+    function createIdentity(
+        string memory name,
+        Metadata memory metadata,
+        address[] memory _owners,
```
codenamejason, line 54: Can we standardize how we write the params, I am fine with either way, i.e.  `_param, param`. Typically I use without _ and only use that for private/internal variables and functions.

0xZakk: I think the most common style is to use the underscore at the beginning of a function parameter (i.e. `_param`) and otherwise to not use them. That would be my preference

thelostone-mc: let's use prefix underscore

contracts/core/Registry.sol
```diff
@@ -23,30 +22,42 @@ contract Registry is AccessControl {
     struct IdentityDetails {
         string name;
         Metadata metadata;
-        address attestationAddr;
+        address anchor;
```
thelostone-mc, line 25: I do prefer `attestationAddress` cause it's more informative and goes without idea of have EAS as first class citizens

codenamejason: How about `attestor`?

thelostone-mc: Hmm this wouldn't be the attestor, this is the subject being attested :P

codenamejason: Ah good point

contracts/core/Registry.sol
zobront: This seems to be the last big open question. I'm not sure either of the solutions above is right.

## CONSTRAINTS 

Here are the constraints we're working with...
1) make it consistent across chains
2) make it unique
3) have it stay unchanged when important values (like name) are unchanged, but reset when those values are changed


## PROBLEMS 

But the solutions above all miss some of these...
- if we include a globally incremented identityId, we can't be consistent across chains
- if we don't include identityId, we aren't necessarily unique (for example, current implementation would let the same user create two identities with the same name, and they'd have the same address)

Additionally, there is the slightly awkward problem that the attestation address is based on the original msg.sender, but that isn't saved, so when we update it we use the new msg.sender ([see here](https://github.com/allo-protocol/allo-v2/blob/da3cf15f6c4f10a85d9bff81319109ac4ddcde1a/contracts/core/Registry.sol#L108). No bueno.

## SOLUTION

The only solution I can think of that solves for all constraints is...
- when an identity is created, we immediately create an `identityId` that is the hash of the `msg.sender` and a `nonce` that is incremented per `msg.sender` (ie it's stored in a mapping and incremented each time they create an identity)
- that `identityId` is saved as a part of the `identity` struct
- the `attestationAddress` (or whatever we choose to call it) is the hash of `identityId` and `name`
- we can NEVER change `identityId`, it's immutable once the identity is created and has nothing to do with ownership, the original deployer has no ownership or rights
- but if we change name, we just use the old `identityId` to generate a new `attestationAddress`

I think this solves for all constraints. Let me know if it's unclear or if we should chat about it more.

contracts/core/Registry.sol
thelostone-mc: anchor works.

contracts/core/Registry.sol
thelostone-mc: I think sticking to anchor works for now. It does feel vague but better than the other options

contracts/core/Registry.sol
thelostone-mc: This means :

- If owner updates name -> use same identityId and generate new attestationAddress
- If owners are updated via grantRoles -> we just revert them and ask them to create a new identity themselves

This feels fair. The only downside is that we can't create a new identity for them cause we won't be able to port their owners and members cause those can be iterated through. That feels like a big issue.

My question: 
Why do we need to create a new identity / generate a new attention address anytime the owner is changed. 
I would say grantRole / revokeRole shouldn't change things in your identity

thelostone-mc: Thinking about a little further, even having a identityId as a counter makes sense  ?
A project with identityId 1 on chain A will be different from identityId 1 on chain B 
We could instead simplify this by saying :
- name cannot be updated 
-  The attestation address/. anchor is generated from `msg.sender`, `owner` and `name`
- And all lookups are by anchor / attestation address.

You can change members without any issues.
If you want to change owners / name -> you would have to create a new identity 

@0xZakk + @nfrgosselin

nfrgosselin: Per our chat on standup, I'm dropping a list of use cases we want to consider for the identity registry.

**most important things**
- allowing projects to be _created_ easily
- allowing projects to build up reputation
- preventing fake projects / impersonating another project (e.g. creating a fake Gitcoin)
- preventing partial owner(s) of a group-owned identity stealing it from the group without consent

**things we want to be possible (even if they have some friction)**
- rebranding a project
- transferring ownership of a project

**things we want to limit**
- creating / selling false identities

contracts/core/Registry.sol
```diff
@@ -204,11 +203,25 @@ contract Registry is AccessControl {
...
-    // ! Who can add / remove owner for an identity? No one. It's a single owner.
-    // ? Transfer ownership will not be supported for now.
-    // - Owner can add / remove members
-    // - Single owner
+    // Note: Check with product if we can retain this function as if a role
+    // can be granted, we should have a way to revoke / transfer ownership
+    /// @notice Transfers the ownership of the identity to a new owner
+    /// @param _identityId The identityId of the identity
+    /// @param _owner New Owner
+    // function changeIdentityOwner(
```
thelostone-mc, line 211: @nfrgosselin curious what your thoughts are

cc @codenamejason

codenamejason: I think this will be okay as we can have a `changeIdentityOwner()` just create a new Identity with the new owner's address and index then generating a new anchor. The old Identity will stay intact and accessible from the original owner and index.

thelostone-mc: Ah so that's the nice part about this. 
Based on the friday call with zak, in the current design :  
- the identityIndex and anchor are not derived from the owner
- changing the owner -> doesn't result in a change in anchor AKA the old identity stays intact

codenamejason: Exactly, we don't ever lose history and is fraud resistant also.

contracts/core/Registry.sol
```diff
@@ -1,47 +1,308 @@
...
+pragma solidity 0.8.19;
 
-// use Solady Roles for ownership of projects: https://github.com/Vectorized/solady/blob/main/src/auth/OwnableRoles.sol
-contract Registry {
-    struct IdentityDetails {
+import {AccessControl} from "../../lib/openzeppelin-contracts/contracts/access/AccessControl.sol";
+
+import {Metadata} from "./libraries/Metadata.sol";
+
+contract Registry is AccessControl {
```
zobront, line 8: Question whether you want to grant some Allo multisig the `DEFAULT_ADMIN_ROLE` so that it can grant or revoke roles in emergencies.

zobront (commented):
This looks SO nice. It's already in such a good spot. Left a few comments that all should be pretty quick to resolve — one larger discussion for call tmrw. But this will be ready to lock within a day or so for sure.

Really impressive work.

contracts/core/Registry.sol
```diff
@@ -1,47 +1,315 @@
...
+            revert INDEX_NOT_AVAILABLE();
+        }
+
+        Identity memory identity = Identity(
+            identityId,
+            _nonce,
+            _name,
+            _metadata,
+            _generateAnchor(identityId, _name)
+        );
```
zobront, line 110: My paranoia but I'd recommend using named assignment here, like 
```
Identity({
    id: identityId,
    ....
});
```
Seen too many projects change the order of variables in the struct later (say, for gas optimizations) and end up screwing themselves.

contracts/core/Registry.sol
```diff
@@ -1,47 +1,315 @@
...
+
+        identitiesById[identityId] = identity;
+        anchorToIdentityId[identity.anchor] = identityId;
+
+        // generate roles
+        bytes32 ownerRole = _generateRole(identityId, RoleType.OWNER);
+        bytes32 memberRole = _generateRole(identityId, RoleType.MEMBER);
```
KurtMerbeth, line 117: we shouls think about storing the roles on the struct
so we don't need to generate them again on every use

contracts/core/Registry.sol
```diff
@@ -1,47 +1,308 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.8.0;
+pragma solidity 0.8.19;
 
-// use Solady Roles for ownership of projects: https://github.com/Vectorized/solady/blob/main/src/auth/OwnableRoles.sol
-contract Registry {
-    struct IdentityDetails {
+import {AccessControl} from "../../lib/openzeppelin-contracts/contracts/access/AccessControl.sol";
```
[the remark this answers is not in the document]
thelostone-mc: Ah thanks for that :D

contracts/core/Registry.sol
```diff
@@ -1,47 +1,308 @@
...
+            revert NO_ACCESS_TO_ROLE();
+        }
+
+        address anchor = _generateAnchor(_identityId, _name);
+
+        Identity storage identity = identitiesById[_identityId];
+        identity.name = _name;
+        identity.anchor = anchor;
+
+        anchorToIdentityId[identity.anchor] = _identityId;
```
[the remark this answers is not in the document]
thelostone-mc: Ah that was intentionally left out cause then if you had an older anchor, you would still be able to find out which identity had created it. Is there a downside of that ?

contracts/core/Registry.sol
thelostone-mc: We could have that set via the constructor ?

contracts/core/Registry.sol
```diff
@@ -1,47 +1,308 @@
...
+        bytes32 indexed identityId,
+        Metadata metadata
+    );
+
+    /// ======================
+    /// ==== Constructor =====
+    /// ======================
+    constructor() {
+        // Empty constructor
+    }
```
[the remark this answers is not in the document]
thelostone-mc: Keeping this to set DEFAULT_ADMIN_ROLE

contracts/core/Registry.sol
```diff
@@ -1,47 +1,308 @@
...
+    /// @param _metadata The new metadata of the identity
+    /// @dev Only owner or member can update metadata
+    function updateIdentityMetadata(
+        bytes32 _identityId,
+        Metadata memory _metadata
+    ) external {
+
+        if (!isMemberOfIdentity(_identityId, msg.sender)) {
+            revert NO_ACCESS_TO_ROLE();
+        }
```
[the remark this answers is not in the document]
thelostone-mc: Ah yes this is what we had agreed on

contracts/core/Registry.sol
```diff
@@ -1,47 +1,308 @@
...
+        identitiesById[identityId] = identity;
+        anchorToIdentityId[identity.anchor] = identityId;
+
+        // generate roles
+        bytes32 ownerRole = _generateRole(identityId, RoleType.OWNER);
+        bytes32 memberRole = _generateRole(identityId, RoleType.MEMBER);
+
+        // assign roles
+        _grantRole(ownerRole, _owner);
+        _grantRole(memberRole, _owner);
```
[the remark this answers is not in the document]
thelostone-mc: Okay that's fair. Less confusion

contracts/core/Registry.sol
```diff
@@ -1,47 +1,308 @@
...
+
+    /// @notice Types of roles assigned to an identity
+    enum RoleType {
+        OWNER,
+        MEMBER
+    }
+
+    /// @notice Struct to hold details of an identity
+    struct Identity {
+        bytes32 id;
```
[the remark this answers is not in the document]
thelostone-mc: It' not needed. You are right

contracts/core/Registry.sol
```diff
@@ -1,47 +1,308 @@
...
+    /// @param _members The members to add
+    /// @dev Only owner can add members
+    function addMembers(
+        bytes32 _identityId,
+        address[] memory _members
+    ) external {
+
+        if (!isOwnerOfIdentity(_identityId, msg.sender)) {
+            revert NO_ACCESS_TO_ROLE();
+        }
```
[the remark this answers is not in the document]
thelostone-mc: making this into a modifier for now. Can explore making it into an internal function for gas optimisation down the line

contracts/core/Registry.sol
thelostone-mc: Okie marking this as resolved as the solution we came with @zobront  on Friday covers our usecases

contracts/core/Registry.sol
thelostone-mc: Marking this as resolved -> cause we are discussing this in newer comment below

contracts/core/Registry.sol
```diff
@@ -1,47 +1,308 @@
...
+
+        if (!hasRole(ownerRole, msg.sender)) {
+            revert NO_ACCESS_TO_ROLE();
+        }
 
+        bytes32 memberRole = _generateRole(_identityId, RoleType.MEMBER);
+
+        _grantRole(ownerRole, _owner);
+        _grantRole(memberRole, _owner);
+        _revokeRole(ownerRole, msg.sender);
```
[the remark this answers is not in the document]
thelostone-mc: Things to clarify :

- Do we want to allow change owner
- Should we support multiple owners 
- Should owner be on Identity struct vs using Roles
- When ownership changes, should we attempt to remove member role as well

contracts/core/Registry.sol
thelostone-mc: I quite like that. 
We have to generate them multiple times otherwise.

contracts/core/Registry.sol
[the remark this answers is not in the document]
thelostone-mc: My intention was to ensure an identity taking part in a pool doesn't end up with issues where their old anchor was being used (think direct grants which have no end date)

contracts/core/Registry.sol
codenamejason: We are going to move the `owner` to the `Identity` struct per our conversation with @zobront this morning.

```javascript
struct Identity {
    uint nonce;
    address owner;
    string name;
    Metadata metadata;
    address anchor;
}
```

contracts/core/Registry.sol
thelostone-mc: Decided not to implement this

contracts/core/Registry.sol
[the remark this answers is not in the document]
thelostone-mc: sticking to what @KurtMerbeth suggested

thelostone-mc (dismissed):

contracts/core/Registry.sol
```diff
@@ -1,47 +1,242 @@
...
+contract Registry is AccessControl {
+    /// @notice Custom errors
+    error NO_ACCESS_TO_ROLE();
+    error NONCE_NOT_AVAILABLE();
+
+    /// @notice Types of roles assigned to an identity
+    enum RoleType {
+        OWNER,
+        MEMBER
+    }
```
zobront, line 17: This can be deleted now.

contracts/core/Registry.sol
```diff
@@ -1,47 +1,242 @@
...
+        string memory _name,
+        Metadata memory _metadata,
+        address _owner,
+        address[] memory _members
+    ) external returns (bytes32) {
+        bytes32 identityId = _generateIdentityId(_nonce);
+
+        if (identitiesById[identityId].nonce != 0) {
+            revert NONCE_NOT_AVAILABLE();
+        }
```
zobront, line 95: Careful using nonce here. What if they use a `nonce` of 0? Then they'd be able to overwrite it.

I'd recommend explicitly checking that `_owner != address(0)` when this is called, and then reverting if `identitiesById[identityId].owner != address(0)`

contracts/core/Registry.sol
```diff
@@ -1,47 +1,242 @@
...
+
+        Identity storage identity = identitiesById[_identityId];
+        identity.name = _name;
+
+        // clear old anchor
+        anchorToIdentityId[identity.anchor] = bytes32(0);
+
+        // set new anchor
+        identity.anchor = anchor;
+        anchorToIdentityId[identity.anchor] = _identityId;
```
zobront, line 143: Since `identity` is in storage, you can save gas by using `anchor` here instead of `identity.anchor`.

contracts/core/Registry.sol
```diff
@@ -1,47 +1,242 @@
...
+    }
 
+    /// @notice Transfers the ownership of the identity to a new owner
+    /// @param _identityId The identityId of the identity
+    /// @param _owner New Owner
+    /// @dev Only owner can transfer ownership.
+    function changeIdentityOwner(bytes32 _identityId, address _owner) external isIdentityOwner(_identityId) {
+        identitiesById[_identityId].owner = _owner;
+        emit IdentityOwnerUpdated(_identityId, _owner);
     }
```
zobront, line 192: We might want to make it a two step ownership transfer? Your call, just always scares me the risk of sending to wrong address. That would require adding a `pendingOwner` field as well and then having it be two calls.

contracts/core/Registry.sol
thelostone-mc: Oh lol

contracts/core/Registry.sol
thelostone-mc: Ah good callout

zobront: I would add an explicit check that `_owner != address(0)` so you are sure this invariant is rock solid.

If not, there's the possibility someone could choose address(0) as an owner (maybe wanting to make it immutable) but not realize that would allow accidentally using the same nonce again and overwriting.

zobront: Actually, you know what? Doing `anchor` for this is better because it'll never be zero, since nobody knows the preimage.

contracts/core/Registry.sol
codenamejason: I like it

thelostone-mc (dismissed):

contracts/core/Registry.sol
```diff
@@ -1,47 +1,263 @@
...
+
+    /// @notice Transfers the ownership of the identity to the pending owner
+    /// @param _identityId The identityId of the identity
+    /// @dev Only pending owner can claim ownership.
+    function changeIdentityOwner(bytes32 _identityId) external {
+        Identity storage identity = identitiesById[_identityId];
+
+        if (identity.pendingOwner == address(0)) {
+            revert NO_PENDING_OWNER();
+        }
```
zobront, line 204: This is covered by checking that `== msg.sender` below.

contracts/core/Registry.sol
```diff
@@ -1,47 +1,263 @@
...
+    {
+        identitiesById[_identityId].pendingOwner = _pendingOwner;
+
+        emit IdentityPendingOwnerUpdated(_identityId, _pendingOwner);
+    }
+
+    /// @notice Transfers the ownership of the identity to the pending owner
+    /// @param _identityId The identityId of the identity
+    /// @dev Only pending owner can claim ownership.
+    function changeIdentityOwner(bytes32 _identityId) external {
```
zobront, line 199: Might name is `acceptIdentityOwnership` for clarity?

zobront (commented):
Added 2 tiny nits below. Ready to merge after that 🚀

zobront (approved):

contracts/core/Registry.sol
zobront: Can remove this type of error above too, since it won't be used.

contracts/core/Registry.sol
thelostone-mc: Ah that's much better

contracts/core/Registry.sol
thelostone-mc: Derp how did i miss that

Issue #5, opened 2023-06-19
Skirting Fees


nfrgosselin:
from @zobront 
Starting a thread for us to discuss the fee skirting issue. The issue is that, since projects choose their own strategies, we can't have Allo protocol fees taken on the strategy, or users can easily skirt it. It needs to live on the core Allo.sol contract, but that poses challenges because when fees are taken may not be consistent across projects.

**CURRENT WORKING SOLUTION**

1) When a pool operator calls `createPool()`, they can add tokens and a fee is taken directly at that point. The total balance with which the pool has been funded is stored in the Allo.sol contract.

2) At any point, the operator can all `fundPool()` and add additional funds, and fees are taken at that point. The total balance with which the pool has been funded is updated in the Allo.sol contract.

3) When users call `allocate()` in a strategy that requires them to send additional funds (ie Quantitative Funding), that can be handled in two ways:
    a) If the funds go directly to the recipient, our prebuilt strategies take the fee directly at that point. If they deploy their own strategy, they could skirt this fee.
    b) If the funds go into the distribution strategy for later distribution, a function is called on Allo.sol that updates the total funding amount for the pool and takes the fee.

4) In order to trigger payouts, we call `finalize()` on Allo.sol. This checks the balance of the distribution strategy and, if it is greater than the stored amount of what it should be, takes any excess balance and transfers it to the Allo treasury. 

**ISSUES**

1) An allocation strategy can be written that sends allocations straight to the recipient and pays no fees.

2) A distribution strategy can be written that skips over the `finalize()` function. We could add a flag to the main contract that is is now `claimable` once `finalize()` is called (which would stop claimers from using the core contract's `claim()` function if their distribution strategy skirts this step) but even that isn't perfect, because it doesn't account for distributions with ongoing phases.

___

from @KurtMerbeth 

Additional Question, since we will have Pools with and without distribution strategies:
Are we taking only fees from the distribution, or also from allocations? @nfrgosselin 

___

from @thelostone-mc 

How would we ensure that the funds aren't sent directly to the DistributionStrategy and bypass the fundPool function ?

___

from @codenamejason 

Can we do this by only allowing the Allocation Strategy to know the Distribution Strategy address or interact with the Distribution Strategy as the only owner? @thelostone-mc

___

from @nfrgosselin 

> Additional Question, since we will have Pools with and without distribution strategies:

Can you say more @KurtMerbeth? What types of pools wouldn't have distribution strategies?

[...]

PR #24, opened 2023-06-24
chore: update Allo.sol

thelostone-mc:


contracts/core/Allo.sol
```diff
@@ -3,28 +3,52 @@ pragma solidity 0.8.19;
...
+
     /**
      * @notice Initializes the contract after an upgrade
      * @dev In future deploys of the implementation, an higher version should be passed to reinitializer
      */
-    function initialize() public reinitializer(1) {}
+    function initialize(address _registry) public reinitializer(1) {
+        registry = Registry(_registry);
+        // ASK: should there be an update function to update registry ? Ownable contract ?
+        // Should this contract be upgradable
```
thelostone-mc, line 40: @codenamejason @zobront @KurtMerbeth

contracts/core/Allo.sol
```diff
@@ -3,28 +3,52 @@ pragma solidity 0.8.19;
...
+    /// @param _poolId The id of the pool
+    /// @dev calls out to the registry to get the identity metadata
+    function getIdentityInfo(
         uint256 _poolId
-    ) external view returns (PoolData memory, string memory) {
-        // Implement the function here
+    ) external view returns (Pool memory pool, string memory identityMetadata) {
+        pool = pools[_poolId];
+        identityMetadata= registry.identities(pool.identityId).metadata;
+        // ASK: why are we returning the pool
```
thelostone-mc, line 51: @zobront

contracts/core/Allo.sol
```diff
@@ -33,57 +57,108 @@ contract Allo is Initializable {
...
-        // todo: return the poolId? what do we want to return here?
-        return _poolId;
+        if (!registry.isMemberOfIdentity(_identityId, msg.sender)) {
+            revert NO_ACCESS_TO_ROLE();
+        }
+        
+        // ASK should we clone _allocationStrategy / _distributionStrategy or just use them directly?
+        // bytes32 salt = keccak256(abi.encodePacked(msg.sender, nonce, "allocationStrategy"));
+        // address clone = ClonesUpgradeable.cloneDeterministic(_allocationStrategy, salt);
+
```
thelostone-mc, line 73: @codenamejason @zobront @KurtMerbeth

contracts/core/Allo.sol
```diff
@@ -33,57 +57,108 @@ contract Allo is Initializable {
...
+        
+        Pool memory pool = Pool(
+            _identityId,
+            _allocationStrategy,
+            _distributionStrategy,
+            _metadata,
+            false
+        );
+
+        // ASK: Does this function also recieve funds ?
```
thelostone-mc, line 82: @zobront

contracts/core/Allo.sol
```diff
@@ -33,57 +57,108 @@ contract Allo is Initializable {
...
     }
 
     // calls voting.generatePayouts() and then uses return data for payout.activatePayouts()
     // permissionless for anyone to call, checks happen within the strategies
     // check to make sure they haven't skrited around fee
-    function finalize(uint _poolId) external {
-        // Implement the function here
+    function finalize(uint _poolId, bytes memory _data) external {
+
+        // ASK: Do we need _data to allow owner to pass custom data ?
```
thelostone-mc, line 137: @zobront not sure if we are still intending to use finalize
If yes -> then do we allow custom data if a distribution wants to use something that is not given by  generatePayouts?

contracts/core/Allo.sol
```diff
@@ -33,57 +57,108 @@ contract Allo is Initializable {
...
+        pools[_poolId].disributionStrategy.distribute(_data, msg.sender);
     }
 
-    // call to distributionStrategy.close()
+    /// @notice Closes the pool
     function closePool(uint _poolId) external {
-        // Implement the function here
+        // pools[_poolId].allocationStrategy.close(); // ASK: Do we need this ?
+        // pools[_poolId].distributionStrategy.close(); // ASK: Do we need this ?
+        pools[_poolId].active = false;
```
thelostone-mc, line 162: @zobront our distributionStrategy interface doesn't have a close function.
Would it make sense to add it do both allocationStrategy and distributionStrategy ?

thelostone-mc (commented):
Left a few open questions

contracts/core/Allo.sol
KurtMerbeth: I'm totally in favor of having the contracts upgradeable. At least at the beginning.

Will we have only one registry, or a list of "trusted sources? (E.g. registry and registry relayer)

contracts/core/Allo.sol
KurtMerbeth: For the allo-strategies we should create clones, since it's cheaper.

For custom strategies I would suggest the same, since they will be re-usable. But I'm also open for direct use

thelostone-mc: Creating clones does seem nicer <3
Okie we'll pair on this on Monday and get it sorted

contracts/core/Allo.sol
thelostone-mc: Ah, good question. 
I want to say 1 registry cause now you create your identityId and same anchor on every chain. Hence the need to support multiple Allo registries ceases to exist (or am I wrong ?)

codenamejason: Does this mean one registry per chain? or do we have a base registry on a single chain and the front-ends just call the registry on mainnet (PGN) no matter what chain they are on?

contracts/core/Allo.sol
codenamejason: Shouldn't we return the `identityId` so they can then look up the identity information? Does this metadata match the same addresses metadata for the same identity in the Registry contract?

```javascript
struct Identity {
  bytes32 id;
  uint index;
  string name;
  Metadata metadata;
  address anchor;
}

struct Pool {
    bytes32 identityId;
    IAllocationStrategy allocationStrategy;
    IDistributionStrategy distributionStrategy;
    Metadata metadata;
    bool active;
}
```

contracts/core/Allo.sol
codenamejason: We could add the two params we need (`_poolId` & `_amount`) as optional to let the user add funds on pool creation and call the `fundPool()` function.

```javascript
function fundPool(uint _poolId, uint _poolAmt) external payable {
    // ASK: what do we do with _poolAmt ?
}

// Call fundPool() from createPool() if they pass any tokens
function createPool(
    bytes32 _identityId,
    address _allocationStrategy,
    address _distributionStrategy,
    Metadata _metadata,
    uint _poolId,
    uint _amount
) external pure returns (uint poolId) {

    if (!registry.isMemberOfIdentity(_identityId, msg.sender)) {
        revert NO_ACCESS_TO_ROLE();
    }

    Pool memory pool = Pool(
        _identityId,
        IAllocationStrategy(_createClone(_allocationStrategy)),
        IDistributionStrategy(_createClone(_distributionStrategy)),
        _metadata,
        false
    );

    // check for tokens passed in and call fundPool() if _amount > 0
    fundPool(_poolId, _amount);

    // ASK: Does this function also recieve funds ?
    // If so -> ETH / ERC20 ?
    // Should we also accept amount as an argument to check ?

    poolId = _poolIndex++;
    pools[poolId] = pool;

    emit PoolCreated(
        poolId,
        _identityId,
        _allocationStrategy,
        _distributionStrategy,
        _metadata
    );
}
```

contracts/core/Allo.sol
thelostone-mc: I think we'll discuss that on the call today but TLDR:
Since it's possible to port your identity (anchor + identityIndex) by controlling the nonce + msg.sender + name
We don't have to worry about whitelisting multiple registries 


>  Does this mean one registry per chain
Yes. If someone wanted to deploy their own duplicate registry, they could do that and our system would still work cause identity can be ported

contracts/core/Allo.sol
thelostone-mc: Yup good callout. Added those

contracts/core/Allo.sol
```diff
@@ -1,89 +1,270 @@
...
+        } else {
+            // ERC20 Token
+            IERC20Upgradeable(_token).safeTransfer(_to, _amount);
+        }
+    }
+
+    // ASK: Should registry be updatable
+    // ASK: Should we allow multiple registries
+    // ASK: Do we need closePool
+    // ASK: If yes -> do we need openPool to set active = true
```
thelostone-mc, line 269: ^ Open questions

contracts/core/Allo.sol
```diff
@@ -1,89 +1,289 @@
...
+        fee = _fee;
+        emit FeeUpdated(fee);
     }
 
-    // call to distributionStrategy.close()
-    function closePool(uint _poolId) external {
-        // Implement the function here
+    /// @notice Create a determenstic clone of of contract
+    /// @param _contract The address of the contract to clone
+    function _createClone(address _contract) internal returns (address clone) {
```
thelostone-mc, line 252: Move into it's own library which would invoke Allo.createPool

contracts/core/Allo.sol
```diff
@@ -1,89 +1,289 @@
...
+
+    /// @notice Struct to hold details of an Pool
+    struct Pool {
+        bytes32 identityId;
+        IAllocationStrategy allocationStrategy;
+        IDistributionStrategy distributionStrategy;
         Metadata metadata;
+        address token;
+        uint256 amount;
         bool active;
```
thelostone-mc, line 29: remove and have strategy implement getter function

contracts/core/Allo.sol
```diff
@@ -1,89 +1,289 @@
...
+    /// ==========================
+
+    /// @notice Incremental Index
+    uint256 private _poolIndex;
+
+    /// @notice Nonce to create salt for clone strategy
+    uint256 private _nonce;
+
+    /// @notice Fee
+    uint256 public fee; // ASK: should this be percentage ?
```
thelostone-mc, line 43: yes it's a percentage

contracts/core/Allo.sol
```diff
@@ -1,89 +1,289 @@
...
+        uint256 _amount,
+        Metadata memory _metadata
+    ) external payable returns (uint256 poolId) {
+        if (!registry.isMemberOfIdentity(_identityId, msg.sender)) {
+            revert NO_ACCESS_TO_ROLE();
+        }
+
+        Pool memory pool = Pool({
+            identityId: _identityId,
+            allocationStrategy: IAllocationStrategy(_createClone(_allocationStrategy)),
```
thelostone-mc, line 137: move cloning logic into the new contract. 
createPool always uses the _allocationStrategy/ _distributionStrategy given to it

contracts/core/Allo.sol
thelostone-mc: remove `closePool` and have active derived from strategies

contracts/core/Registry.sol
```diff
@@ -22,6 +22,8 @@ contract Registry is AccessControl {
         uint nonce;
         string name;
         Metadata metadata;
+        address owner;
+        address pendingOwner;
```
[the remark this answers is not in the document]
codenamejason: Yeah, it feels kind of weird being there.

contracts/core/Allo.sol
codenamejason: 1. Should registry be updatable - Open
2. Should we allow multiple registries - Open
3. Do we need closePool - Removed `closePool()`
4. If yes -> do we need openPool to set active = true - Removed the `active` property from `Pool` struct

contracts/core/Allo.sol
codenamejason: Latest PR reflects these changes.

Note: @zobront @thelostone-mc @kurtmerbeth how do we want to handle funding the pool?
        1. Do we expose a fundPool() function on the interface?
        2. We still do the transfer to the treasury in the core contract.
        3. Pool transfers happen in the strategies.
       
I know we have ongoing conversation on this here: #5 

@zobront @KurtMerbeth @thelostone-mc WDYT?

zobront: `fundPool()` in core that takes a fee and sends the funds on to the distribution strategy

contracts/core/Registry.sol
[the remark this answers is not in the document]
zobront: I've seen this as the pattern with a number of protocols but also recognize that it's not core info. Separate mapping should work just fine.

contracts/core/Allo.sol
```diff
@@ -1,79 +1,284 @@
...
+    /// @param _data encoded data unique to the allocation strategy for that pool
+    function applyToPool(uint256 _poolId, bytes memory _data) external payable returns (uint256 applicationId) {
+        IAllocationStrategy allocationStrategy = pools[_poolId].allocationStrategy;
+
+        // todo: we are returning bytes here as defined in the interface, should we return a uint256?
+        // Note: @zobront @thelostone-mc @kurtmerbeth
+        bytes memory data = allocationStrategy.applyToPool(_data, msg.sender);
+        applicationId = abi.decode(data, (uint256));
+
+        return applicationId;
```
codenamejason, line 161: What should we return here @zobront @thelostone-mc @KurtMerbeth

contracts/core/Allo.sol
```diff
@@ -1,79 +1,284 @@
...
+        poolId = _poolIndex++;
+        pools[poolId] = pool;
+
+        emit PoolCreated(poolId, _identityId, _allocationStrategy, _distributionStrategy, _token, _amount, _metadata);
+    }
+
+    /// @notice passes _data through to the allocation strategy for that pool
+    /// @param _poolId id of the pool
+    /// @param _data encoded data unique to the allocation strategy for that pool
+    function applyToPool(uint256 _poolId, bytes memory _data) external payable returns (uint256 applicationId) {
```
codenamejason, line 153: What should we return here @zobront @thelostone-mc @KurtMerbeth

KurtMerbeth: it should return the `applicationId` (received from `allocation.applyToPool`) i guess

but according to the `IAllocationStrategy` interface `bytes` are returned. We should think aboutreturning the `applicationId`

contracts/core/Allo.sol
thelostone-mc: Decided that this is the only contract which we'll make upgradeable. Later on once we are stable -> we'll set the update perms to address(0) and freeze upgradability.

contracts/core/Allo.sol
thelostone-mc: - Added function to make registry updatable  
- Nope we stick to one registry (based on yesterdays calls)

contracts/core/Allo.sol
thelostone-mc: @zobront @codenamejason @KurtMerbeth So i believe the direction we wanted to head was 
- Allo.sol would not invoke _createClone. Instead always use the address
- We would move clone logic into another library 

This is where I get a little fuzzy. The options discussed where: 
- We would have an additional contract `StrategyClone.sol` which exposes 1 function which would leverage Clone library to clone contract and then invoke Allo.createPool 
OR
- Where we have a whitelisted mapping of string(bytes) -> address in Allo. If you pass that string -> we'll do a look up and if we find a mapping -> clone , else not clone 
OR
- We just have createPoolWithClone and createPool in Allo.sol and leave it at that

KurtMerbeth: > Where we have a whitelisted mapping of string(bytes) -> address in Allo. If you pass that string -> we'll do a look up and if we find a mapping -> clone , else not clone

we should avoid going this path because it will be super confusing
I'm in favor of clone all, or if you decide against, we should have two functions where `createPool` checks if it's a whitelisted strategy or not. if yes -> `_createPoolWithClone`. Just to be safe

KurtMerbeth: Can we talk about edge cases of third party strategies?

**Scenario 1:**
In the first scenario, there are two pools:

In Pool 1, a third-party deploys a cloneable strategy that is not included in our whitelisted list. The pool is initialized using a clone of that strategy. 
In Pool 2, the pool is initialized using the implementation of the cloneable strategy directly, without making a clone.


**Scenario 2:**
In the second scenario, there are again two (or more) pools:
In Pool 1, a third-party strategy is deployed and the pool is initialized directly using that implementation without creating a clone. 
In Pool 2, the pool is also initialized with the same strategy, at the same address, again without creating.

This situation can potentially cause issues if the third-party strategy is poorly designed. Since both pools are using the same implementation at the same address, any problems or vulnerabilities in the strategy can affect both pools simultaneously.

contracts/core/Registry.sol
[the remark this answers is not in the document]
thelostone-mc: Fixed via https://github.com/allo-protocol/allo-v2/pull/30

contracts/core/Allo.sol
thelostone-mc: > I'm in favor of clone all, or if you decide against, we should have two functions where createPool checks if it's a whitelisted strategy or not. if yes -> _createPoolWithClone. Just to be safe

This makes sense. Easier to implement and our mapping adds a safety check
Personally the more I think about , the more it makes sense to always clone

contracts/core/Allo.sol
codenamejason: Pairing with @thelostone-mc we came up with this:

```javascript
function applyToPool(uint256 _poolId, bytes memory _data) external payable returns (bytes memory) {
    IAllocationStrategy allocationStrategy = pools[_poolId].allocationStrategy;

    return allocationStrategy.applyToPool(_data, msg.sender);
}
```

thelostone-mc: synced with @codenamejason. We are returning bytes cause we don't know if the applicationId would always be a unit or if an allocation strategy wanted to send a different kind of data.

KurtMerbeth: @thelostone-mc 
any ideas of what other data a strategy might return?

contracts/core/Allo.sol
codenamejason: I am assuming we should have the whitelist of strategies onchain. I added a new contract called `Whitelist.sol` and a function in `Allo.sol` to call/check the addresses of the strategies the user is trying to deploy.

`Whitelist.sol` contains the following functions:
```javascript
mapping(address => bool) public whitelist;

constructor() {}

function addToWhitelist(address _address) external {
    whitelist[_address] = true;
}

function removeFromWhitelist(address _address) external {
    whitelist[_address] = false;
}

function isWhitelisted(address _address) external view returns (bool) {
    return whitelist[_address];
}
```

I didn't add any role management and they are wide open currently but didn't want to go all in if this was not a path we wanted to pursue.

contracts/core/Allo.sol
thelostone-mc: paired with @KurtMerbeth. making this return uint256.

@zobront

contracts/core/Allo.sol
thelostone-mc: Open question :
- should this live within allo.sol
- have 2 functions createPool and createPoolWithClone 

@zobront

contracts/core/Allo.sol
```diff
@@ -1,79 +1,283 @@
...
+            identityId: _identityId,
+            allocationStrategy: IAllocationStrategy(Clone.createClone(_allocationStrategy, _nonce++)),
+            distributionStrategy: IDistributionStrategy(Clone.createClone(_distributionStrategy, _nonce++)),
+            metadata: _metadata
+        });
+
+        poolId = _poolIndex++;
+
+        if (_amount > 0) {
+            // TODO: Ask product, should we have minimum fee
```
thelostone-mc, line 155: @codenamejason

contracts/core/Allo.sol
thelostone-mc: We've rewritten this to defer clonability to the strategy 
aa93a6d0b087741c0c142293bc4c815a23039237 

Curious what you think
@zobront + @0xZakk

zobront: I am in favor of it being a separate contract but it doesn't have to be. The important thing is that it's separated into two functions.

`createPool()` => takes in addresses for strategies and uses them, no clones
`createPoolWithClone()` => uses takes in strategies from approved list, clones them, then calls `createPool()` using those clones as arguments, and marks it as "approved" because strategies came from library

You know I like consolidating into fewer contracts but I am in favor of these being separate because it keeps the logic of Allo.sol clean (it uses the addresses its passed as strategies) and makes the StrategyLibrary.sol a peripheral contract that Allo owns, can be easily updated, can track all the approved strategies, etc. 

But your call, this is fine either way.

contracts/core/Allo.sol
zobront: @thelostone-mc Agreed, returning a `uint` is fine, let's change the strategy interface to reflect that.

contracts/core/Allo.sol
nfrgosselin: Documenting what we discussed in standup today:

We want to have two fee settings on the contract, which will both be controllable via Gitcoin DAO governance votes:
- **protocol fee** — this should be a percentage that is taken out of the pool amount. E.g. a 2% fee on a 100 DAI pool would yield 2 DAI to the protocol fee and 98 DAI to be distributed via the pool. The owner would need to deposit a total of 100 DAI into the pool.
- **minimum protocol fee** — this should be a flat amount. If the owner creates a pool where the calculated protocol fee is less than this minimum amount, then they will be required to pay this minimum fee.

Both of these values should be set to zero to start.

zobront: @thelostone-mc The only thing I can think of is requiring the min fee to be sent in ETH in addition to the funding amount. So it would be "base fee" as opposed to "min fee", and you'd probably just keep it at 0.

contracts/core/Allo.sol
```diff
@@ -1,79 +1,359 @@
...
     }
 
-    // transfers _poolAmt from msg.sender to the pool, and takes a fee
-    function fundPool(uint256 _poolId, uint256 _poolAmt) external payable {
-        // Implement the function here
+    /// @notice Fund a pool
+    /// @param _poolId id of the pool
+    /// @param _amount extra amount of the token to be deposited into the pool
+    /// @param _token The address of the token that the pool is denominated in
+    // TODO: Ask product, should we restrict to pool admin?
```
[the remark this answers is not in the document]
zobront: I don't see a need to.

contracts/core/libraries/Clone.sol
```diff
@@ -0,0 +1,29 @@
...
+        return codeSize > 0;
+    }
+
+    /// @notice Create a determenstic clone of of contract
+    /// @param _contract The address of the contract to clone
+    /// @param _nonce The nonce to use for the clone
+    function createClone(address _contract, uint256 _nonce) internal returns (address clone) {
+        require(_isContract(_contract), "not a contract");
+        bytes32 salt = keccak256(abi.encodePacked(msg.sender, _nonce++));
+        clone = ClonesUpgradeable.cloneDeterministic(_contract, salt);
```
zobront, line 25: What's the reasoning to use `cloneDeterministic` instead of `clone`? 

That would save needing a nonce at all, and not sure what they goal is because you can't actually trust the determinism, since another use could increment the nonce before you.

If determinism is important, than you'll want a separate nonce per users to make sure it's reliable.

contracts/core/Allo.sol
```diff
@@ -1,79 +1,359 @@
...
+    /// ====================================
+    /// =========== Modifier ===============
+    /// ====================================
+
+    modifier isPoolAdmin(uint256 _poolId) {
+        if (!registry.isOwnerOrMemberOfIdentity(pools[_poolId].identityId, msg.sender)) {
+            revert NO_ACCESS_TO_ROLE();
+        }
+        _;
+    }
```
zobront, line 120: Did we decide we don't want any ownership on Pools and just want to trust that all members and owners on Registry are able to control the pool?

I think that's where we landed, but want to make sure, as it'll likely mean that "identities" will need to fragment into subidentities to manage permissions properly.

contracts/core/Allo.sol
```diff
@@ -1,79 +1,359 @@
...
+    /// ====================================
+    /// ==== External/Public Functions =====
+    /// ====================================
+
+    /// @notice Returns the pool info
+    /// @param _poolId The id of the pool
+    /// @dev calls out to the registry to get the identity metadata
+    function getPoolInfo(uint256 _poolId) external view returns (Pool memory) {
+        return pools[_poolId];
+    }
```
zobront, line 131: I don't think this is needed, since we have the `pools` public getter.

contracts/core/Allo.sol
```diff
@@ -1,79 +1,359 @@
...
+        address _allocationStrategy,
+        address payable _distributionStrategy,
+        address _token,
+        uint256 _amount,
+        Metadata memory _metadata,
+        bool _cloneAllocationStrategy,
+        bool _cloneDistributionStrategy
+    ) external payable returns (uint256 poolId) {
+        /// Check the boolean flag to see if the strategy should be cloned
+        /// Additionally, force clone the strategy if it's approved
```
zobront, line 153: This comment is no longer accurate.

contracts/core/Allo.sol
```diff
@@ -1,79 +1,359 @@
...
+        address allocationStrategy =
+            _cloneAllocationStrategy ? Clone.createClone(_allocationStrategy, _nonce++) : _allocationStrategy;
+
+        address distributionStrategy =
+            _cloneDistributionStrategy ? Clone.createClone(_distributionStrategy, _nonce++) : _distributionStrategy;
```
zobront, line 158: Do we want to check `_isApprovedStrategy()` here? I think this would be the best place.

contracts/core/Allo.sol
```diff
@@ -1,79 +1,359 @@
...
+        usedStrategies[_distributionStrategy] = true;
+
+        Pool memory pool = Pool({
+            identityId: _identityId,
+            allocationStrategy: IAllocationStrategy(_allocationStrategy),
+            distributionStrategy: IDistributionStrategy(_distributionStrategy),
+            metadata: _metadata
+        });
+
+        poolId = _poolIndex++;
```
zobront, line 214: This is a weird one but this will start at `poolId = 0`. This can sometimes cause issues because you might want to check `if (poolId == 0)` somewhere as proof that it doesn't exist.

I would recommend using `++poolIndex` to start at 1, or manually incrementing it in the constructor / initializer to make sure it starts at 1.

contracts/core/Allo.sol
```diff
@@ -1,79 +1,359 @@
...
-    // validates that the owner is actually allowed to use the identity
-    function createPool(PoolData memory /*_poolData*/ ) external pure returns (uint256) {
-        uint32 _poolId = 0;
+        if (_amount > 0) {
+            _fundPool(_token, _amount, poolId, address(pool.distributionStrategy));
+        }
 
-        // todo: return the poolId? what do we want to return here?
-        return _poolId;
+        pools[poolId] = pool;
```
zobront, line 220: I haven't thought through this completely but this doesn't follow `checks-effects-interactions` because `_fundPool` is reenterable (external call of ETH to untrusted address). 

Could lead to reentrancy risk where someone creates a pool, then calls back to themselves and can create another pool they shouldn't be allowed to because `pools[poolId]` hasn't been set with that info yet?

I'd flip the order of these to be safe.

contracts/core/Allo.sol
```diff
@@ -1,79 +1,359 @@
...
+        uint256 _amount,
+        Metadata memory _metadata
+    ) internal returns (uint256 poolId) {
+        if (!registry.isOwnerOrMemberOfIdentity(_identityId, msg.sender)) {
+            revert NO_ACCESS_TO_ROLE();
+        }
+
+        if (usedStrategies[_allocationStrategy] || usedStrategies[_distributionStrategy]) {
+            revert STRATEGY_ALREADY_USED();
+        }
```
zobront, line 202: Hmmm the risk here is that someone could frontrun your transaction and use your strategies to DOS you. Is there some data that will always be stored on the strategy that could be checked here to make sure it's the right user calling it?

Once we're needing to do that, we might just want to make strategies verify that they are unused themselves. Like, for example, enforce an `initialize()` function as a part of this call with proper validation to make sure it has to be used by the right person (ie `require(msg.sender == Allo.sol && caller == owner());` or something)

Sorry for back and forth on this. Hadn't thought about the DOS risk until I saw it.

Any ideas on how to do this check elegantly in a way that strategy deployers won't screw up?

contracts/core/Allo.sol
codenamejason: This checks that the user is either a member or owner of an `Identity` not a `Pool`.

contracts/core/Allo.sol
codenamejason: Agreed, we can remove this.

contracts/core/Allo.sol
nfrgosselin: I think it makes sense for either a member or owner to be able to control the pool. 

@zobront can you say more about this part:
> as it'll likely mean that "identities" will need to fragment into subidentities to manage permissions properly.

contracts/core/Allo.sol
codenamejason: Do we want to do something like this?

```javascript
if (!_isApprovedStrategy(_allocationStrategy) && !_isApprovedStrategy(_distributionStrategy)) {
    revert NOT_APPROVED_STRATEGY();
}

address allocationStrategy =
      _cloneAllocationStrategy ? Clone.createClone(_allocationStrategy, _nonce++) : _allocationStrategy;

address distributionStrategy =
  _cloneDistributionStrategy ? Clone.createClone(_distributionStrategy, _nonce++) : _distributionStrategy;
```
or
```javascript
address allocationStrategy =
    _cloneAllocationStrategy && _isApprovedStrategy(_allocationStrategy)
    ? Clone.createClone(_allocationStrategy, _nonce++)
    : _allocationStrategy;

address distributionStrategy = _cloneDistributionStrategy && _isApprovedStrategy(_distributionStrategy)
    ? Clone.createClone(_distributionStrategy, _nonce++)
    : _distributionStrategy;
```

What about if they wanted to create a clone of their own, unapproved contract?

contracts/core/Allo.sol
```diff
@@ -1,79 +1,359 @@
...
+        if (_feePercentage > FEE_DENOMINATOR) {
+            revert INVALID_FEE_PERCENTAGE();
+        }
+
+        feePercentage = _feePercentage;
+
+        emit FeeUpdated(feePercentage);
+    }
+
+    /// @notice Add a strategy to the whitelist
```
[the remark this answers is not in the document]
codenamejason: Yes, we already spoke about this and have updated it.

contracts/core/Allo.sol
zobront: @nfrgosselin Imagine there is an Allo identity. You create a pool and show that you are a part of Allo (ie use Allo as the identity for the pool). That means every single member of Allo is an admin of your pool, but you may not want that.

In other words, there may be people that you want to give permission to create a pool as a part of Allo, but not permission to be an admin for all Allo pools.

The result is that you'd need to have an "Allo Admin" identity, etc to split it more granularly.

contracts/core/Allo.sol
zobront: They can't create a clone of their own unapproved contract on Allo.sol. They'd need to do that themselves. 

Clones on this contract exist for cloning our approved contracts, nothing else!

zobront: It would be more like:
```
if (_cloneAllocationStrategy) {
    require(_isApprovedStrategy(_allocationStrategy));
    ...
} 

if (_cloneDistributionStrategy) {
    require(...)
}
```

codenamejason: Should we use `require()` or custom error?

```javascript
if (!_isApprovedStrategy(_allocationStrategy)) {
    revert STRATEGY_NOT_APPROVED();
}
```

`require` is a simpler and more straightforward approach, especially when the error message doesn't require detailed information. Custom error handling, on the other hand, allows us to provide more context-specific error messages, which can be helpful for debugging/testing and providing meaningful feedback to our users.

What are your thoughts?

zobront: I'm just giving you pseudocode of the idea of where checks should happen. Stay consistent across the contract with custom errors for sure.

codenamejason: Understood, thank you 🙏

contracts/core/Allo.sol
codenamejason: Good call-out @zobront 

For our reference:
The CEI pattern prevents reentrancy attacks by ensuring that all state changes are made before any interactions with other contracts occur. This is done by dividing the code of a function into three phases:

Checks: This phase performs all necessary checks on the input parameters and the current state of the contract.
Effects: This phase makes any changes to the state of the contract.
Interactions: This phase makes calls to other contracts.

By following this pattern, it is guaranteed that any state changes will be committed before any interactions with other contracts occur. This prevents malicious contracts from calling a function in another contract and then calling that function again before the first call has finished executing.

contracts/core/libraries/Clone.sol
codenamejason: Is there a reason we need to determine the address for the strategies @thelostone-mc @KurtMerbeth

thelostone-mc: Nah we could just use clones.

thelostone-mc: fixed via ce70255

contracts/core/Allo.sol
thelostone-mc: Hmm might be worth discussing this on a call cause we'd def want multiple users to be able to operate the round. If we can;t use members, the only option is identity owner but that's a single address / contract which would have to be changed (or we add in a new layer of privileged_members :P )

contracts/core/Allo.sol
thelostone-mc: > They can't create a clone of their own unapproved contract on Allo.sol. They'd need to do that themselves.

> It would be more like:
> 
> ```
> if (_cloneAllocationStrategy) {
>     require(_isApprovedStrategy(_allocationStrategy));
>     ...
> } 
> 
> if (_cloneDistributionStrategy) {
>     require(...)
> }
> ```

This might not work cause if we would want to allow a user to clone an approved allocation strategy but use their custom distribution strategy.

Given the requirements: 
- We want to allow cloning of ONLY approved contracts
- We want to allow people to one approved strategy (clone) + one custom strategy 

Here is a simpler way we can do this ad2c6d6

```javascript
    function createPoolWithClone(
        bytes32 _identityId,
        address _allocationStrategy,
        address payable _distributionStrategy,
        address _token,
        uint256 _amount,
        Metadata memory _metadata
    ) external payable returns (uint256 poolId) {
        address allocationStrategy =
            _isApprovedStrategy(_allocationStrategy) ? Clone.createClone(_allocationStrategy) : _allocationStrategy;
        address distributionStrategy = _isApprovedStrategy(_distributionStrategy)
            ? Clone.createClone(_distributionStrategy)
            : _distributionStrategy;

        return _createPool(_identityId, allocationStrategy, payable(distributionStrategy), _token, _amount, _metadata);
    }
    ```

File: contracts/core/Allo.sol (hunk 23 above)
Comment (thelostone-mc) [reply contracts/core/Allo.sol]: Looks like this is taken care of. Just checking

File: contracts/core/Allo.sol (hunk 24 above)
Comment (thelostone-mc) [reply contracts/core/Allo.sol:187]: > Once we're needing to do that, we might just want to make strategies verify that they are unused themselves. Like, for example, enforce an initialize() function as a part of this call with proper validation to make sure it has to be used by the right person (ie require(msg.sender == Allo.sol && caller == owner()); or something)

Is this the same as :
- Interfaces now become an abstract contract
- Having modifiers to offer those checks   `require(msg.sender == Allo.sol && caller == owner());`

File: contracts/core/Allo.sol (hunk 18 above)
Comment (zobront) [reply contracts/core/Allo.sol]: Yeah. I think this might be the best option. The only other thing I could see would be to have roles on Allo.sol itself based on PoolID for who is allowed to manage each pool, and identity isn't assumed to be used for admin purposes except setting up new pools.

File: contracts/core/Allo.sol (hunk 24 above)
Comment (zobront) [reply contracts/core/Allo.sol:187]: Did we decide that we will enforce all pool creators having identities? I think yes, right?

If that's the case, this solution seems right to me.

File: contracts/core/libraries/Clone.sol (hunk 17 above)
Comment (zobront) [reply contracts/core/libraries/Clone.sol]: @KurtMerbeth That Sherlock issue is mostly factually incorrect. If there was a reorg, then the transaction to send funds would be wound back as well.

However, I do see the point where there might be some risk of it being funded separate from the initial create but within 15-30 seconds, lining up with a reorg, having the funding txs be replayable, and then having an attacker create a new pool with the old address and running the replayable funding tx.

Might be safer to go back to the old way. Seems almost impossible to happen but cost of tracking the nonces is pretty small, so maybe can't hurt.

File: contracts/core/Allo.sol (hunk 23 above)
Comment (codenamejason) [reply contracts/core/Allo.sol]: Updating.

```javascript
poolId = ++_poolIndex;
pools[poolId] = pool;

if (_amount > 0) {
  _fundPool(_token, _amount, poolId, address(pool.distributionStrategy));
  emit PoolCreated(poolId, _identityId, _allocationStrategy, _distributionStrategy, _token, _amount, _metadata);
}

// forcing 0 for the amount bc that either means the tx to transfer failed or they didn't send any funds
emit PoolCreated(poolId, _identityId, _allocationStrategy, _distributionStrategy, _token, 0, _metadata);
```

contracts/core/Allo.sol
codenamejason: Yes, I agree. We had that in there this morning and adding it back now.

contracts/core/Allo.sol
KurtMerbeth: tbh i like the idea of having `members` and `members with privilege` @thelostone-mc

contracts/core/libraries/Clone.sol
thelostone-mc: Hmm alright. Given that we wanted to go the deterministic way :
7d5ce5b

contracts/core/Allo.sol
```diff
@@ -1,79 +1,368 @@
...
+
+    /// @notice Fee percentage
+    /// @dev 1e18 = 100%, 1e17 = 10%, 1e16 = 1%, 1e15 = 0.1%
+    uint256 public feePercentage;
+
+    /// @notice Incremental index
+    uint256 private _poolIndex;
+
+    /// @notice Nonce for cloning strategies
+    uint256 private _nonce;
```
zobront, line 47: Make this a mapping by address. The reason is that that will guarantee that a user can know their deployment address in advance. It's unlikely to be used but might be handy for a front end or whatever to show.

The current set up is probably fine, but having a deterministic address that could be changed by someone else using it the same block is an unexpected behavior it's better to avoid.

contracts/core/Allo.sol
```diff
@@ -1,79 +1,368 @@
...
+    address payable public treasury;
+
+    /// @notice Registry of pool creators
+    Registry public registry;
+
+    /// @notice Pool.id -> Pool
+    mapping(uint256 => Pool) public pools;
+
+    /// @notice Strategy -> bool
+    mapping(address => bool) public approvedStrategies;
```
zobront, line 59: I'd probably call this `strategyLibrary` for clarity that other strategies are fine, these are just the ones we provide.

contracts/core/Allo.sol
```diff
@@ -1,79 +1,368 @@
...
+
+        registry = Registry(_registry);
+        treasury = _treasury;
+        feePercentage = _feePercentage;
+
+        emit RegistryUpdated(_registry);
+        emit TreasuryUpdated(_treasury);
+        emit FeeUpdated(_feePercentage);
+
+        __Multicall_init();
```
zobront, line 110: I need another bit of time to think through if there are any safety implications of multicall here. I think it's fine.

contracts/core/Allo.sol
```diff
@@ -1,79 +1,368 @@
...
+    }
+
+    /// ====================================
+    /// =========== Modifier ===============
+    /// ====================================
+
+    modifier onlyPoolAdmin(uint256 _poolId) {
+        if (!registry.isOwnerOrMemberOfIdentity(pools[_poolId].identityId, msg.sender)) {
+            revert UNAUTHORIZED();
+        }
```
zobront, line 120: I know this is still an open discussion. I'm leaning that it would be better to store role data for each pool separately rather than assuming same as identity, but don't feel super strongly. It's more of a product question. Let's talk about it on the call.

zobront: Currently this is only used for changing Metadata, but keep in mind it will be called from strategies to control everything — approving applications, submitting merkle root for claims, etc.

contracts/core/Allo.sol
zobront: Reopening this conversation. I think @KurtMerbeth has the right solution, but doesn't look like it's been implemented.

We would need a check here that `_identityId == strategy.identityId()` for both strategies.

While we're doing that, it might make more sense to:
- Get rid of the `usedStrategies` mapping
- Call the `initialize()` function on the strategies straight from the pool
- In that function, check that `initialized = false` (which accomplishes same as `usedStrategies`
- Any non clone strategy will, in that function, check that `_identityId (argument) == identityId (locally stored)` (same as Kurt's recommendation, where `identityId` should be set in constructor
- Sets `intialized = true` and whatever else is needed

This means we should probably also add a generic `bytes memory` field when creating a pool that gets passed as initialize data.

contracts/core/Allo.sol
```diff
@@ -1,79 +1,368 @@
...
+
+    /// @notice passes _data through to the allocation strategy for that pool
+    /// @notice returns the applicationId from the allocation strategy
+    /// @param _poolId id of the pool
+    /// @param _data encoded data unique to the allocation strategy for that pool
     function applyToPool(uint256 _poolId, bytes memory _data) external payable returns (uint256) {
-        // Implement the function here
+        IAllocationStrategy allocationStrategy = pools[_poolId].allocationStrategy;
+
+        return allocationStrategy.applyToPool(_data, msg.sender);
```
zobront, line 239: Can make this one line to match `distribute` and `allocation`.

contracts/core/Allo.sol
```diff
@@ -1,79 +1,368 @@
...
+        if (_token == address(0)) {
+            // Native Token
+            (bool sent,) = _to.call{value: _amount}("");
+            if (!sent) {
+                revert TRANSFER_FAILED();
+            }
+        } else {
+            // ERC20 Token
+            IERC20Upgradeable(_token).transfer(_to, _amount);
+        }
```
zobront, line 360: I recommend using `safeTransferLib` here. They handle all the weird ERC20 edge cases where things break. Check out this list to see how wonky it can get: https://github.com/d-xo/weird-erc20

contracts/core/Allo.sol
thelostone-mc: i think we could rely on a indexing service ! making it iterable feels like an overkill

contracts/core/Allo.sol
thelostone-mc: makes sense. sounds good to me

contracts/core/Allo.sol
thelostone-mc: Agreed this is alright over the call

contracts/core/Allo.sol
thelostone-mc: As per discussion -> we'd handle pool permission by having roles for pool admin via OZ roles

contracts/interfaces/IAllocationStrategy.sol
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 pragma solidity 0.8.19;
 
+import "@openzeppelin-upgradeable/proxy/utils/Initializable.sol";
```
thelostone-mc, line 4: Do we need this ?

KurtMerbeth: No, i think we just forgot to remove it.

contracts/core/Allo.sol
thelostone-mc: fixed via 7f8ce4ad29ea7aede1a61ea0a7cc3f381b6ad061

contracts/core/Allo.sol
thelostone-mc: fixed via 0eb884141e1bb0eeb745aff14cd96342330aa891

contracts/core/Allo.sol
thelostone-mc: fixed via 1fffea3

contracts/core/Allo.sol
thelostone-mc: fixed via 327d43ceb2806c59aad6dead10a9d4f2affb961a

contracts/core/Allo.sol
thelostone-mc: fixed via 48eda1dca29b059ff39ac58bdd351b815d44ac55

thelostone-mc: fixed via 48eda1dca29b059ff39ac58bdd351b815d44ac55

contracts/core/Allo.sol
thelostone-mc: fixed via 7f8ce4ad29ea7aede1a61ea0a7cc3f381b6ad061

contracts/core/Allo.sol
thelostone-mc: Fixed via 1aaaf57

contracts/core/Allo.sol
thelostone-mc: This is now updated to allow pools to have a set of owners which is set by the identity owner who created the pool

contracts/core/Allo.sol
```diff
@@ -96,26 +102,30 @@ contract Allo is Initializable, Ownable, MulticallUpgradeable {
...
     }
 
     /// ====================================
     /// =========== Modifier ===============
     /// ====================================
 
-    modifier onlyPoolAdmin(uint256 _poolId) {
-        if (!registry.isOwnerOrMemberOfIdentity(pools[_poolId].identityId, msg.sender)) {
+    modifier onlyPoolOwner(uint256 _poolId) {
+        if (!hasRole(keccak256(abi.encodePacked(_poolId)), msg.sender)) {
```
zobront, line 128: Is there a reason to make this the hash of the pool ID instead of the pool ID itself? It seems nice for it to be legible.

contracts/core/Allo.sol
```diff
@@ -1,79 +1,460 @@
...
+                i++;
+            }
+        }
 
-        // todo: return the poolId? what do we want to return here?
-        return _poolId;
+        // grant admin roles to pool creator
+        _grantRole(POOL_ADMIN_ROLE, msg.sender);
+        // set admin role for POOL_OWNER_ROLE
+        _setRoleAdmin(POOL_OWNER_ROLE, POOL_ADMIN_ROLE);
```
zobront, line 288: Personally I think I'd name these opposite? It doesn't really matter but I think of admins as all the people who can control the pool, and the owner is the one who actually owns it and has ultimate say.

contracts/core/Allo.sol
```diff
@@ -1,79 +1,460 @@
...
+        uint256 _amount,
+        Metadata memory _metadata,
+        address[] memory _owners
+    ) external payable returns (uint256 poolId) {
+        address allocationStrategy;
+        address distributionStrategy;
+
+        // DISCUSS: We either
+        // - allow cloning of every contract (use the bool)
+        // - or we force cloning of only approved contracts
```
zobront, line 168: Is this DISCUSS still relevant or is it old? I think current implementation here is perfect, but if there's still question, let's talk about it.

contracts/core/Allo.sol
codenamejason: You thinking something like this:

```javascript
// grant pool admin roles
for (uint256 i = 0; i < _admins.length;) {
    _grantRole(POOL_ADMIN_ROLE, _admins[i]);
    unchecked {
        i++;
    }
}

// grant owner role to pool creator
_grantRole(POOL_OWNER_ROLE, msg.sender);

// Note: do we want this? set admin role for pool creator
// _grantRole(POOL_ADMIN_ROLE, msg.sender);
```

contracts/core/Allo.sol
thelostone-mc: > Roles are referred to by their bytes32 identifier. These should be exposed in the external API and be unique. The best way to achieve this is by using public constant hash digests:

Isn't that cause the role has to bytes32 
https://docs.openzeppelin.com/contracts/3.x/api/access#AccessControl

contracts/core/Allo.sol
thelostone-mc: > Owners and administrators both have all permissions, including posting listings, editing the organization's profile, and managing the permissions of other administrators, but an owner has control over other owners as well as administrators.

LOL derp you're right @zobront  :P Will change that 
Admin are the folks who control the pool while owners can control who is an admin / not

thelostone-mc: Changes via 487ab9d0f9dc334037401c498461016e67737814

contracts/core/Allo.sol
thelostone-mc: This still needs to be discussed cause the boolean arguments don't make a lot of sense at this point the way the contract is written. Looking at the code you realise, that the bool is in theory a dummy argument cause :

- if you pass bool as true but the strategy is not approved -> it reverts
- if you pass bool as true but the strategy is approved -> it clones
- if you pass bool as false but the strategy is approved -> it reverts
- if you pass bool as false but the strategy is not approved -> it doesn't clone 

This feels the  the same as 
- if contract is approved -> it clones
- if contract Is not approved -> it doesn't clone

The boolean is added just to share more clarity on reverts which I'm not sure if worth it (i could be wrong but hence the discussion )

thelostone-mc:
@zobront Merging this off so that we can proceed with the strategies PR 
Things to discuss :

- https://github.com/allo-protocol/allo-v2/pull/24#discussion_r1254954588
-

Things to verify :
- https://github.com/allo-protocol/allo-v2/pull/24#discussion_r1255259356
- https://github.com/allo-protocol/allo-v2/pull/24#discussion_r1255258252

PR #27, opened 2023-06-27
chore: remove hardhat

thelostone-mc:
### Description

- remove hardhat
- add storage check via foundry
- update toml file to include build info 
- update .env.sample
- regenerate yarn.lock file


codenamejason (approved):
🔥

PR #30, opened 2023-06-27
chore: move pendingOwner to mapping

### LINKED ISSUE #26: can we move pendingOwner into a separate mapping?
can we move pendingOwner into a separate mapping?
just because it's not an important identity information and there is no need to pass this information when you quey an identity

_Originally posted by @KurtMerbeth in https://github.com/allo-protocol/allo-v2/pull/24#discussion_r1242564964_

thelostone-mc:
fixes #26

### DIFF
--- a/contracts/core/Registry.sol
+++ b/contracts/core/Registry.sol
@@ -211,7 +211,7 @@ contract Registry is AccessControl {
         }
 
         identity.owner = newOwner;
-        identityIdToPendingOwner[_identityId] = address(0);
+        delete identityIdToPendingOwner[_identityId];
 
         emit IdentityOwnerUpdated(_identityId, identity.owner);
     }
--- a/contracts/core/Registry.sol
+++ b/contracts/core/Registry.sol
@@ -17,7 +17,6 @@ contract Registry is AccessControl {
         string name;
         Metadata metadata;
         address owner;
-        address pendingOwner;
         address anchor;
     }
 
@@ -30,6 +29,9 @@ contract Registry is AccessControl {
 
     /// @notice anchor -> Identity.id
     mapping(address => bytes32) public anchorToIdentityId;
+
+    /// @notice Identity.id -> pending owner
+    mapping(bytes32 => address) public identityIdToPendingOwner;
 
     /// ======================
     /// ======= Events =======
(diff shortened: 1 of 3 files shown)


KurtMerbeth (approved):

codenamejason (approved):

codenamejason (approved):

contracts/core/Registry.sol
```diff
@@ -71,6 +73,12 @@ contract Registry is AccessControl {
         return identitiesById[identityId];
     }
 
+    /// @notice Retrieve identity's pending owner
+    /// @param identityId The identityId of the identity
+    function getIdentityPendingOwner(bytes32 identityId) public view returns (address) {
+        return identityIdToPendingOwner[identityId];
+    }
+
```
zobront, line 81: I don't think you need this getter honestly. Already have the mapping public, and it's not something that'll really be accessed.

contracts/core/Registry.sol
```diff
@@ -197,13 +204,14 @@ contract Registry is AccessControl {
...
 
-        if (msg.sender != identity.pendingOwner) {
+        if (msg.sender != newOwner) {
             revert NOT_PENDING_OWNER();
         }
 
-        identity.owner = identity.pendingOwner;
-        identity.pendingOwner = address(0);
+        identity.owner = newOwner;
+        identityIdToPendingOwner[_identityId] = address(0);
```
zobront, line 214: I usually use `delete identityIdToPendingOwner[_identityId]` but not sure if it's equivalent gas wise.

zobront (dismissed):
Looks good to merge. Made two small optional comments.

contracts/core/Registry.sol
codenamejason: When you use delete, it clears the storage slot associated with the given key, effectively removing the key-value pair from the mapping. The `delete` operation refunds the gas cost of the storage slot deletion. On the other hand, when you assign address(0) to a mapping value, it actually writes the value address(0) to the storage slot, consuming gas for the storage write operation.

KurtMerbeth (dismissed):

test/foundry/core/Registry.t.sol
```diff
@@ -270,7 +270,8 @@ contract RegistryTest is Test {
 
         assertFalse(registry.isOwnerOfIdentity(newIdentityId, owner), "after: notAnOwner");
         assertTrue(registry.isOwnerOfIdentity(newIdentityId, notAMember), "after: isOwner");
-        assertEq(registry.getIdentityPendingOwner(newIdentityId), address(0), "after: pendingOwner");
+        // Todo: mock this another way
+        // assertEq(pendingOwner, address(0), "after: pendingOwner");
```
thelostone-mc, line 274: ^ @codenamejason We might want to fix this test

thelostone-mc: Fixed this via 5603a4feb5bbfaa2d01b3879db36ad7761e636d8

PR #31, opened 2023-06-27
adding offchain qf spec

nfrgosselin:


contracts/strategies/allocation/linear-qf/qf-offchain-spec.md
```diff
@@ -0,0 +1,60 @@
...
+
+At time of writing (June 2023), most prominent QF formulas are too computationally expensive to be calculated on-chain at a reasonable cost. As a step towards an eventual fully on-chain QF solution, this strategy functions a hybrid solution: 
+* votes are recorded **on-chain**
+* the distribution is calculated **off-chain**
+* the final distribution is recorded **on-chain**
+
+In order to facilitate this, the contract will emit vote events so that off-chain calculation services can pull in the relevant data. 
+
+## User flows
+Full flow is documented here: https://miro.com/app/board/uXjVMXyfa-o=/?share_link_id=797514162800
```
0xZakk, line 15: Can you export this as a png and drop it here in the doc?

0xZakk (commented):
I think this is probably good! Will let others weight in too.

We could definitely make it more technically detailed, but this is probably good enough for now

contracts/strategies/allocation/linear-qf/qf-offchain-spec.md
nfrgosselin: Good call, added

codenamejason (commented):
This looks really good! I think we could definitely start building something from this and raise questions/feedback.

Off-chain side note: we would just need to PR to update the indexer or one of their team do it (calculations). The Graph is straightforward.

Our options for statuses:
- From Allo v1: bitmap solution
- Use an `enum`:
```javascript
enum ApplicationStatus {
    NONE,
    PENDING,
    APPROVED,
    REJECTED
}
```
We could also add more if we need to.

For the Sybil defense settings we should write a list of what we think we should be supporting.

thelostone-mc:
Could we just link it to the miro doc so that updates becomes easier ? As opposed to swapping out png

contracts/strategies/allocation/linear-qf/qf-offchain-spec.md
```diff
@@ -0,0 +1,61 @@
...
+## User flows
+Full flow is shown here: 
+![Alt text](<Strategy Evaluation - QF Vote with events.png>)
+
+### Pool Creation
+- identity owner or member creates pool via `createPool` method on `Allo.sol`
+- core QF settings are recorded on allocation strategy contract, including
+    - Application start/end and voting start/end dates
+    - On-chain application requirements 
+    - Sybil Defense settings (minimum contribution, matching cap, donor matching eligibility requirements)
```
thelostone-mc, line 23: Would these be on-chain or part of the metadata ?
min contribution feels like on-chain but the others would be part of metadata

contracts/strategies/allocation/linear-qf/qf-offchain-spec.md
```diff
@@ -0,0 +1,61 @@
...
+- applicant applies to pool via `applyToPool` on `Allo.sol`
+    - application only processed if application period is open
+    - otherwise, revert
+- allocation strategy checks that application is eligible
+    - if yes, application marked as `Pending`
+    - if no, application marked as `Rejected`
+
+### Application approval
+- pool owner or member makes decision via `reviewApplications` on allocation strategy contract
+-
```
thelostone-mc, line 36: might be worth calling out that
- applications would have to valid entries from the registry 
- every identity can apply a max of 1 time
- the application status can however be toggled

contracts/strategies/allocation/linear-qf/qf-offchain-spec.md
```diff
@@ -0,0 +1,61 @@
...
+    - Sybil Defense settings (minimum contribution, matching cap, donor matching eligibility requirements)
+    - Pool metadata
+
+### Application
+- applicant applies to pool via `applyToPool` on `Allo.sol`
+    - application only processed if application period is open
+    - otherwise, revert
+- allocation strategy checks that application is eligible
+    - if yes, application marked as `Pending`
+    - if no, application marked as `Rejected`
```
thelostone-mc, line 32: If we want to stick to every identity can apply a max of 1 time. 
It would be worth adding the logic where status could be Pending / reapplied ?

nfrgosselin:
> 

@0xZakk Can you say more about the technical details you would want to see? Want to start making this a template for us

contracts/strategies/allocation/linear-qf/qf-offchain-spec.md
nfrgosselin: > applications would have to valid entries from the registry

I don't think that's true — we decided that the protocol only requires registry identities for creating a pool, right? It's up to the pool owner to decide whether or not they want to require a valid registry entry for application. So that feels like a setting that we could create during pool creation?

Makes sense on the other two points though 🙌

contracts/strategies/allocation/linear-qf/qf-offchain-spec.md
nfrgosselin: I think it's a mix of on-chain and off-chain metadata. My feeling is that the settings should only be on-chain if there's logic that would be used in the contract — everything else would be metadata. 

So I think the list would be:
- On-chain: application start/end, voting start/end, require registry identity True/False
- Off-chain: minimum contribution, matching cap, donor requirements

wdyt?

thelostone-mc: LGTM.

thelostone-mc:
@nfrgosselin could we create a new folder called spec and throw all these in there as opposed to having it within the contract folder ?

contracts/strategies/allocation/linear-qf/qf-offchain-spec.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Application
+- applicant applies to pool via `applyToPool` on `Allo.sol`
+- strategy checks that application is valid:
+    - Is application period open?
+        - If no, revert with message that application period is closed
+    - If Allo registry identity required, does the application have valid Allo registry identity?
+        - If no, revert with message that pool requires a valid Allo registry identity
+- If application is valid, mark as `Pending`
+
```
KurtMerbeth, line 49: are we allowing to create multiple applications by one identity?
@nfrgosselin

contracts/strategies/allocation/linear-qf/qf-offchain-spec.md
```diff
@@ -0,0 +1,85 @@
...
+    - If application is valid (see decision tree above), then proceed to following checks
+    - If the original application's local status is `Pending`, then the application is reverted with the message that application has already been submitted
+    - If the original application's local status is `Rejected`, then set application's status as `Re-applied`
+
+
+### Donation voting
+- donor submits donation via `allocate` on `Allo.sol`
+    - donation only processed if voting period is open
+    - donations can be sent in native tokens or ERC20s
+    - donation(s) immediately distributed to recipient
```
KurtMerbeth, line 66: we should avoid having funds directly distributed to the recipient

the reason is simple:
i could deploy a contract as receipient and would be able to flashloan big donations or loop through donations. Basically i would be able to loop $10 through the allocate function and generate a total allocation of millions of dollars

KurtMerbeth: > donations can be sent in native tokens or ERC20s

so an allocation can be in any token the donor wants to send?

thelostone-mc (approved):

contracts/strategies/allocation/linear-qf/qf-offchain-spec.md
thelostone-mc: Nope! That could be a different strategy if needed

PR #36, opened 2023-06-28
Add strategy - QF Offchain calculations

codenamejason:
This PR builds out a basic allocation strategy using off-chain calculations for applicants.


contracts/strategies/allocation/QFOffchainCalc.sol
```diff
@@ -0,0 +1,194 @@
...
+        REJECTED,
+        REAPPLIED
+    }
+
+    /// @notice recipient -> Status -> ApplicationStatus
+    mapping(address => mapping(Status => ApplicationStatus)) public globalStatusMap;
+
+    /// @notice recipient -> Application
+    mapping(address => Application) public applications;
+
```
KurtMerbeth, line 61: we can't use `recipient -> XYZ` mappings here, since different projects/applications can have the same receipient.
maybe we can use `applicationId` instead

contracts/strategies/allocation/QFOffchainCalc.sol
```diff
@@ -0,0 +1,194 @@
...
+    /// @notice applicationId -> Application
+    mapping(uint256 => Application) public applicationIdToApplication;
+
+    constructor(address _alloCoreContract) {
```
KurtMerbeth, line 65: since the strategy will be cloneable, we have to remove the constructor and use initialize instead

q: we don't want the initialization to happen during the pool creation?

contracts/strategies/allocation/QFOffchainCalc.sol
```diff
@@ -0,0 +1,194 @@
...
+        }
+
+        if (identityRequired) {
+            // Todo: check that the user has an identity and revert if not
+            // revert IDENTITY_REQUIRED();
+        }
+
+        Application memory application = Application({
+            applicationId: ++_index,
+            recipient: sender,
```
KurtMerbeth, line 88: the sender is not the recipient. recipient should be part of the `_data`

contracts/strategies/allocation/QFOffchainCalc.sol
```diff
@@ -0,0 +1,194 @@
...
+    /// @notice Returns the status of the application
+    /// @param applicationId The applicationId of the application
+    function getApplicationStatus(uint256 applicationId) external view override returns (ApplicationStatus) {
+        return globalStatusMap[msg.sender][applicationIdToApplication[applicationId].status];
+    }
+
+    /// @notice Allocates votes to the application(s)
+    /// @param _data The data to be decoded
+    /// @param sender The sender of the transaction
+    function allocate(bytes memory _data, address sender) external payable override onlyOwner returns (uint256) {
```
KurtMerbeth, line 123: q: shouldn't we always pass a `_data` array here?

contracts/strategies/allocation/QFOffchainCalc.sol
```diff
@@ -0,0 +1,194 @@
...
+    /// ========= Custom Functions =========
+    /// ====================================
+
+    /// @notice Creates a pool
+    /// @param _identityId The identityId of the pool
+    /// @param _distributionStrategy The distribution strategy of the pool
+    /// @param _token The token of the pool
+    /// @param _amount The amount of the pool
+    /// @param _metadata The metadata of the pool
+    function createPool(
```
KurtMerbeth, line 156: This function is not needed since the creation happens on Allo directly, or?

contracts/strategies/allocation/QFOffchainCalc.sol
KurtMerbeth: We are not using the status bitmap anymore?

contracts/strategies/allocation/QFOffchainCalc.sol
```diff
@@ -0,0 +1,194 @@
...
+        address indexed voter, uint256 indexed applicationId, uint256 vote, address token, address recipient
+    );
+
+    /// @notice Struct to hold details of an application
+    struct Application {
+        address recipient;
+        uint256 applicationId;
+        uint256 amount;
+        uint256 applicationTime;
+        uint256 votingTime;
```
nfrgosselin, line 42: why do we need a votingTime value on the application?

contracts/strategies/allocation/QFOffchainCalc.sol
```diff
@@ -0,0 +1,194 @@
...
+        uint256 _amount,
+        Metadata memory _metadata
+    ) external onlyOwner {
+        allo.createPool(_identityId, address(this), payable(_distributionStrategy), _token, _amount, _metadata);
+    }
+
+    /// @notice Review applications
+    /// @dev Pool admin can only set applications to `ACCEPTED`, `REJECTED` or `REAPPLIED`
+    function reviewApplications() external onlyOwner {
+        // Todo: Pool admin can only set applications to `ACCEPTED`, `REJECTED` or `REAPPLIED`
```
nfrgosselin, line 169: The pool admin shouldn't be able to set the "reapplied" status — that should happen automatically when a project reapplies

contracts/strategies/allocation/QFOffchainCalc.sol
```diff
@@ -0,0 +1,194 @@
...
+            votingTime: 0,
+            votes: 0,
+            status: Status.PENDING
+        });
+
+        // Note: not sure where to grab the poolId here?
+        // allo.applyToPool(_poolId, _data);
+
+        // Set the status's of the application
+        globalStatusMap[sender][Status.PENDING] = IAllocationStrategy.ApplicationStatus.Pending;
```
nfrgosselin, line 100: Might be missing this somewhere, but I don't see a local status map. How are you thinking these will play together?

contracts/strategies/allocation/QFOffchainCalc.sol
```diff
@@ -0,0 +1,194 @@
...
+
+import {IAllocationStrategy} from "../../interfaces/IAllocationStrategy.sol";
+import {Allo} from "../../core/Allo.sol";
+import {Metadata} from "../../core/libraries/Metadata.sol";
+
+/// @title QFOffchainCalc
+/// @notice A strategy that allows users to apply to a pool and vote on applications
+/// @dev This strategy is used for QF pools that have offchain calculations
+/// @author allo-team
+contract QFOffchainCalc is IAllocationStrategy, Ownable {
```
thelostone-mc, line 14: nitpick: need a better name :P

contracts/strategies/allocation/QFOffchainCalc.sol
thelostone-mc: I don't think we should use status bitmap

contracts/strategies/allocation/QFOffchainCalc.sol
```diff
@@ -0,0 +1,190 @@
...
+        if (msg.sender != address(allo)) {
+            revert NOT_ALLO();
+        }
+        _;
+    }
+
+    /// @notice Initialize the contract
+    /// @param _allo The address of the Allo contract
+    /// @param _owner The address of the owner
+    function initialize(address _allo, address _owner) public initializer {
```
codenamejason, line 73: @zobront Would it make sense to have this callable from `Allo.sol` on create and have every strategy implement this.
This would mean we pass custom bytes data to allow init to work as expected based on the strategy.

zobront: Yep, I think that's a great idea.

codenamejason: I added to a recent PR - Let me know what you think.

contracts/strategies/allocation/QFOffchainCalc.sol
codenamejason: How is this? `BaseAllocationWithVoting`

contracts/strategies/allocation/AllocationWithVotingBase.sol
```diff
@@ -0,0 +1,306 @@
...
+
+        (uint256 applicationId, address recipient, bytes32 identityId, Metadata memory metadata) =
+            abi.decode(_data, (uint256, address, bytes32, Metadata));
+
+        if (identityRequired && !registry.isOwnerOrMemberOfIdentity(identityId, _sender)) {
+            revert UNAUTHORIZED();
+        }
+
+        // Note: applicationId is 0 when the application is new
+        if (applicationId == 0) {
```
KurtMerbeth, line 117: Should we add a check, if there is already a application for the identityId?
With check: only one application by one identity
Without check: many applications by one identity

 [spec](https://github.com/allo-protocol/allo-v2/pull/31/files#r1247589613)

@nfrgosselin

contracts/strategies/allocation/AllocationWithVotingBase.sol
```diff
@@ -0,0 +1,306 @@
...
+        if (payouts.length > 1) {
+            // Transfer funds
+            // Note: this should live in distribution strategy
+            // bool transferSuccess = Transfer.transferPayouts(token, payouts);
+            // if (!transferSuccess) {
+            //     revert Transfer.TRANSFER_FAILED();
+            // }
+
+            // Loop through the payouts and update the votes on each application
+            for (uint256 i; i < payouts.length; i++) {
```
KurtMerbeth, line 189: i think there is a problem in the logic

for **every** receipient in payouts we iterate through **all** applicationIds and allocate the amounts
my suggestion: we shouldnt use `Payout` here or we should change payout to
```
Payout {
   uint256 applicationId;
   uint256 amount;
}
```

in general i would suggest a format like:

```
struct SingleAlloction {
   uint256 applicationId;
   uint256 amount;
}

struct Allocation {
   address token;
   uint256 totalAmount;
   SingleAllocation[] allocations;
}
```

for every `token` in the list, we would have one transfer of `totalAmount` and than we could iterate over the `allocations` and add the amounts.
if we do not allow to allocate different token we could remove the outer struct.
if we go with multiple token, we need to introduce a `token -> amount` mapping for each user

from spec:
> donations can be sent in native tokens or ERC20s

contracts/strategies/allocation/AllocationWithVotingBase.sol
```diff
@@ -0,0 +1,306 @@
...
+        } else {
+            // Transfer funds
+            // Note: this should live in distribution strategy
+            // bool transferSuccess = Transfer.transferPayout(token, payouts[0]);
+            // if (!transferSuccess) {
+            //     revert Transfer.TRANSFER_FAILED();
+            // }
+
+            Application storage application;
+            application = applications[applicationIds[0]];
```
KurtMerbeth, line 218: redundant code block

contracts/strategies/allocation/AllocationWithVotingBase.sol
```diff
@@ -0,0 +1,306 @@
...
+            emit ApplicationVoted(
+                sender, applications[applicationIds[0]].applicationId, payouts[0].amount, token, application.recipient
+            );
+            // Note: Return the total votes for Application?
+            return 1;
+        }
+    }
+
+    /// @notice Generates the payouts for the application(s)
+    function generatePayouts() external payable override onlyOwner returns (bytes memory) {}
```
KurtMerbeth, line 230: `generatePayouts` is supposed to return an array of recipients and allocations/percentages
therefore we need to be able to iterate through the (approved) applications

q: What's the most gas friendly way to implement it? How to deal with hitting the block gas limit if the list is too long?
@zobront

contracts/strategies/allocation/AllocationWithVotingBase.sol
```diff
@@ -0,0 +1,306 @@
...
+    /// @dev Pool admin can only set applications to `ACCEPTED` or `REJECTED`
+    function reviewApplications(bytes memory _data) external {
+        // Todo: Pool admin can only set applications to `ACCEPTED` or `REJECTED`
+        (bytes32 identityId,,,) = allo.pools(poolId);
+
+        if (!registry.isOwnerOfIdentity(identityId, msg.sender)) {
+            revert UNAUTHORIZED();
+        }
+
+        // Note: How to determine single vs multi application calls?
```
KurtMerbeth, line 247: we could expect `_data` to always be an array
single review -> array of length 1.
so we could always iterate through the array

contracts/interfaces/IAllocationStrategy.sol
```diff
@@ -15,8 +15,11 @@ interface IAllocationStrategy {
         Rejected
     }
 
+    // Initializer
+    function initialize(bytes memory _data) external;
```
thelostone-mc, line 19: @zobront Does it make sense to have an init function ?

contracts/interfaces/IAllocationStrategy.sol
```diff
@@ -15,8 +15,11 @@ interface IAllocationStrategy {
...
+
     // call to allo() to get identity for pool, then to registry() to get metadata
-    function getOwnerIdentity() external view returns (string memory);
+    function getOwnerIdentity() external view returns (bytes32);
```
thelostone-mc, line 22: @zobront updating this to return the identityId of the pool creator

contracts/strategies/allocation/AllocationWithVotingBase.sol
thelostone-mc: We should add a check !

contracts/strategies/allocation/AllocationWithVotingBase.sol
thelostone-mc: No we will allowing only 1 token per transaction.
That token can be an ERC20 / native token

contracts/interfaces/IDistributionStrategy.sol
```diff
@@ -4,6 +4,13 @@ pragma solidity 0.8.19;
 import "./IStrategy.sol";
 
 interface IDistributionStrategy is IStrategy {
+    /// @notice initialize the strategy
+    /// @param _identityId the identityId of the pool
+    /// @param _poolId the poolId of the pool
+    /// @param _allo the allo address
+    /// @param _data the data to initialize the strategy
+    function initialize(bytes32 _identityId, uint256 _poolId, address _allo, bytes memory _data) external;
```
thelostone-mc, line 12: This should be removed cause it's already defined in IStrategy

contracts/strategies/allocation/AllocationWithVotingBase.sol
```diff
@@ -116,44 +116,53 @@ contract AllocationWithVotingBase is IAllocationStrategy, Ownable, Initializable
     }
 
     modifier onlyPoolAdmin() {
-        if (!registry.isOwnerOrMemberOfIdentity(allo.pools[_poolId].identityId, msg.sender)) {
+        (bytes32 identityId,,,) = allo.pools(poolId);
```
thelostone-mc, line 119: This isn't right. Owners are now stored on the Pool.
So we'd have to rely on the OZ roles instead defined in Allo.sol

contracts/strategies/allocation/AllocationWithVotingBase.sol
```diff
@@ -116,44 +116,53 @@ contract AllocationWithVotingBase is IAllocationStrategy, Ownable, Initializable
...
+        if (!registry.isOwnerOrMemberOfIdentity(identityId, msg.sender)) {
             revert UNAUTHORIZED();
         }
         _;
     }
 
     /// @notice Constructor
-    function AllocationWithVotingBase(bytes32 _identityId_, address _allo) {
+    function allocationWithVotingBaseConstructor(bytes32 _identityId_, address _allo) public {
```
thelostone-mc, line 127: we could likely rename as this constructor.

contracts/strategies/allocation/AllocationWithVotingBase.sol
```diff
@@ -176,14 +185,13 @@ contract AllocationWithVotingBase is IAllocationStrategy, Ownable, Initializable
...
-                metadata: metadata,
-                creator: _sender
-            });
+            application = applications[applicationId];
+            application.identityId = identityId;
+            application.applicationId = applicationId;
+            application.recipient = recipient;
+            application.status = Status.PENDING;
+            application.metadata = metadata;
+            application.creator = _sender;
```
thelostone-mc, line 194: ^ could we revert to the older format. Just to keep it consistent.
readability also is nicer on the eye in the older format

contracts/strategies/allocation/AllocationWithVotingBase.sol
```diff
@@ -203,7 +211,8 @@ contract AllocationWithVotingBase is IAllocationStrategy, Ownable, Initializable
         }
 
         // Add the application to the applications mapping
-        applications[applicationId] = application;
+        // todo: fix next line
+        // applications[applicationId] = application;
```
thelostone-mc, line 215: not clear on why we are commenting this out 🤔

contracts/strategies/allocation/AllocationWithVotingBase.sol
codenamejason: cant when using mapping inside.

contracts/strategies/allocation/AllocationWithVotingBase.sol
```diff
@@ -226,28 +235,41 @@ contract AllocationWithVotingBase is IAllocationStrategy, Ownable, Initializable
         }
 
         // Decode the _data
-        Allocations[] memory allocationsPerToken = abi.decode(_data, Allocations[]);
+        (Allocations[] memory allocationsPerToken) = abi.decode(_data, (Allocations[]));
```
thelostone-mc, line 238: Do we need the `()`

contracts/strategies/allocation/AllocationWithVotingBase.sol
codenamejason: the assignment fails with the nested mapping.

contracts/strategies/allocation/AllocationWithVotingBase.sol
codenamejason: yes, it returns a tuple.

contracts/strategies/allocation/AllocationWithVotingBase.sol
```diff
@@ -226,28 +235,41 @@ contract AllocationWithVotingBase is IAllocationStrategy, Ownable, Initializable
...
+            // convert to TransferData
+            Transfer.TransferData memory transferData =
+                Transfer.TransferData({to: address(this), amount: allocationsTotalAmount});
+
             // transfer total amount of token to this contract
-            Transfer.transferAmount(allocations.token, allocationsTotalAmount);
+            Transfer.transferAmount(allocations.token, transferData);
 
-            for (uint256 i = 0; i < allocations.data.length;) {
+            for (uint256 j = 0; j < allocations.data.length;) {
```
thelostone-mc, line 252: Ah nice catch

contracts/strategies/allocation/AllocationWithVotingBase.sol
```diff
@@ -226,28 +235,41 @@ contract AllocationWithVotingBase is IAllocationStrategy, Ownable, Initializable
...
-        Allocations[] memory allocationsPerToken = abi.decode(_data, Allocations[]);
+        (Allocations[] memory allocationsPerToken) = abi.decode(_data, (Allocations[]));
         for (uint256 i; i < allocationsPerToken.length;) {
             // Get the allocations per token
             Allocations memory allocations = allocationsPerToken[i];
 
             uint256 allocationsTotalAmount = allocations.totalAmount;
 
+            // convert to TransferData
+            Transfer.TransferData memory transferData =
```
thelostone-mc, line 246: not clear why we are converting this here.  Do we need this ?

contracts/strategies/allocation/AllocationWithVotingBase.sol
```diff
@@ -226,28 +235,41 @@ contract AllocationWithVotingBase is IAllocationStrategy, Ownable, Initializable
...
+                allocationsTotalAmount -= allocations.data[j].amount;
+
+                // TOOD: check
+                emit ApplicationVoted(
+                    sender,
+                    allocations.data[j].applicationId,
+                    allocations.data[j].amount,
+                    allocations.token,
+                    application.recipient
+                );
```
thelostone-mc, line 269: ^ we don't want to emit a vote for every vote. This gets quite expensive 
 Let's revert to the older format.

contracts/strategies/allocation/AllocationWithVotingBase.sol
codenamejason: Yes, the `Transfer.tranferAmounts()` takes a `TransferData` struct.

`function transferAmount(address _token, TransferData memory _transferData)`
`function transferAmounts(address _token, TransferData[] memory _transferData)`

```javacript
struct Allocations {
    address token;
    uint256 totalAmount;
    Allocation[] data;
}

struct Allocation {
    uint256 applicationId;
    uint256 amount;
}

struct TransferData {
    address to;
    uint256 amount;
}
```

We either send an array of `TransferData` or a single `TransferData` object.

contracts/strategies/allocation/AllocationWithVotingBase.sol
codenamejason: I moved it outside the `for` loop and added a comment. Also added a new event for multiple at the same time. We can either keep both or eliminate the singular event.

` emit ApplicationsVoted(sender, new uint256[](0), new uint256[](0), new address[](0));`

This still needs to be finished. I am not sure yet how to derive the data to pass to the event.

contracts/interfaces/IAllocationStrategy.sol
codenamejason: @zobront for reference we moved two functions to `IStrategy.sol`

```javascript
interface IStrategy {
    /**
     * STORAGE (with public getters)
     *     bool initialized;
     *     bytes32 identityId;
     *     uint256 poolId;
     *     address allo;
     */

    // initialize the strategy with the poolId and allo address
    // set initialized to true and ensure it can't be called again
    // check if identityId passed, is same as the identityId set during deployment
    // if identityId is not set during deployment, then set it (for clones)
    function initialize(bytes32 _identityId, uint256 _poolId, address _allo, bytes memory _data) external;

    // call to allo() to return identity for pool
    function getOwnerIdentity() external view returns (bytes32);
}
```

contracts/strategies/allocation/AllocationWithVotingBase.sol
codenamejason: I was also assuming we would force the `_data` to be an array of `Allocations`.

contracts/strategies/allocation/AllocationWithVotingBase.sol
codenamejason: Not sure if this is still the case, some updates have been made since and don't see it.

contracts/strategies/allocation/AllocationWithVotingBase.sol
thelostone-mc: Ah makes sense

contracts/strategies/allocation/AllocationWithVotingBase.sol
thelostone-mc: It doesn't look like this is resolved

contracts/strategies/allocation/AllocationWithVotingBase.sol
KurtMerbeth: it's already fixed @codenamejason

contracts/strategies/allocation/AllocationWithVotingBase.sol
thelostone-mc: Gotcha

contracts/strategies/allocation/AllocationWithVotingBase.sol
codenamejason: I never pushed (face slap) and you updated 👍🏼

contracts/strategies/allocation/AllocationWithVotingBase.sol
codenamejason: How do you mean?

contracts/core/Allo.sol
```diff
@@ -249,44 +258,47 @@ contract Allo is Initializable, Ownable, AccessControl {
...
+        _grantRole(POOL_ADMIN_ROLE, msg.sender);
+        // set admin role for POOL_MANAGER_ROLE
+        _setRoleAdmin(POOL_MANAGER_ROLE, POOL_ADMIN_ROLE);
+
         // initialize strategies
         // @dev Initialization is expect to revert when invoked more than once
-        _allocationStrategy.initialize(_identityId, poolId, address(this), _initAllocationData);
-        _distributionStrategy.initialize(_identityId, poolId, address(this), _initDistributionData);
+        _allocationStrategy.initialize(address(this), _identityId, poolId, _initAllocationData);
+        _distributionStrategy.initialize(address(this), _identityId, poolId, _token, _initDistributionData);
```
KurtMerbeth, line 292: @thelostone-mc @codenamejason @zobront 

what do you think about adding some sanity checks here?
I think about:
`_allocationStrategy.identityId == _identityId`, `_allocationStrategy.poolId == _poolId`, `_allocationStrategy.allo == address(this)` etc..

just to ensure, that the used strategies are follwing somehow our standards.

codenamejason: Not a bad idea but our interfaces don't expose that data. The only way would be to enforce the `identityId` & `poolId` be exposed somewhere Allo can access on the strategies that are being initialized. Am I thinking about this correctly?

Screenshot 2023-07-10 at 09 44 22

contracts/core/Allo.sol
```diff
@@ -340,6 +344,10 @@ contract Allo is Initializable, Ownable, AccessControl {
             revert NOT_ENOUGH_FUNDS();
         }
 
+        if (_token == address(0) && msg.value != _amount) {
+            revert AMOUNT_MISMATCH();
+        }
```
thelostone-mc: ^ ~is it worth adding a check for ERC20 balance as well ?~
 Maybe not :P

contracts/strategies/allocation/allocation-offchain-calculation/AllocationWithOffchainCalculations.sol
```diff
@@ -25,6 +25,16 @@ contract AllocationWithOffchainCalculations is BaseAllocationStrategy, Transfer,
...
 
+    // Note: this is mapped to the Allo global status's in the mapping below.
+    /// @notice Enum for the local status of the recipient
+    enum Status {
+        NONE,
+        PENDING,
+        ACCEPTED,
+        REJECTED,
+        REAPPLIED
+    }
```
thelostone-mc, line 36: Do we need want to maintain this at the interface level ?

contracts/strategies/allocation/allocation-gating/AllocationGating.sol
```diff
@@ -91,6 +91,11 @@ abstract contract AllocationGating is BaseAllocationStrategy {
         }
     }
 
+    /// @notice Set ready for payout
+    function setReadyForPayout(bool _ready) external onlyPoolManager {
+        payoutReady = _ready;
+    }
```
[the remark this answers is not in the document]
KurtMerbeth: yea. you're right.

contracts/core/libraries/Transfer.sol
```diff
@@ -0,0 +1,94 @@
...
+            revert TRANSFER_FAILED();
+        }
+        if (_token == address(0)) {
+            _transferNative(_to, _amount);
+        } else {
+            // ERC20 Token
+            IERC20(_token).safeTransfer(_to, _amount);
+        }
+    }
+}
```
zobront, line 94: Please use SafeTransferLib instead of implementing this yourself!

contracts/core/Allo.sol
```diff
@@ -124,8 +133,19 @@ contract Allo is Initializable, Ownable, AccessControl {
...
-    modifier onlyPoolOwner(uint256 _poolId) {
-        if (!hasRole(keccak256(abi.encodePacked(_poolId)), msg.sender)) {
+    // NOTE: We have two onlyPoolManager modifiers in the same contract, one without parameters and one with a parameter (_poolId). The other one
+    // is in the BaseStrategy.sol contract. This is not allowed in Solidity. Caught this trying to use the protocol in a demo.
+    // I appended WithPoolId to the modifier name to make it unique. Happy to rename to what the team decides.
+    modifier onlyPoolManagerWithPoolId(uint256 _poolId) virtual {
+        if (!_isPoolManager(_poolId, msg.sender)) {
+            revert UNAUTHORIZED();
+        }
+        _;
```
zobront, line 143: Huh? I don't see any problem with this, since they aren't in the same contract. What am I missing?

contracts/strategies/allocation/BaseAllocationStrategy.sol
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.19;
+
+import "../BaseStrategy.sol";
+import "../../interfaces/IAllocationStrategy.sol";
+
+abstract contract BaseAllocationStrategy is BaseStrategy, IAllocationStrategy {
```
zobront, line 7: Why do we need this contract?

contracts/core/Allo.sol
```diff
@@ -423,32 +538,26 @@ contract Allo is Initializable, Ownable, AccessControl {
     function _fundPool(address _token, uint256 _amount, uint256 _poolId, IDistributionStrategy _distributionStrategy)
         internal
     {
```
zobront, line 540: It feels a lot cleaner for this function to take the fee first (or skip if it's 0) and then distribute, rather than using the if else and recreating the funding logic in both clauses. But just personal preference.

contracts/core/Allo.sol
codenamejason: Was trying to use our contracts in an example and had issues with duplicate names in different inherited contracts:
```javascript
/// @author @codenamejason <jax@jaxcoder.xyz>
contract StrategyRunner is Allo, TokenBalanceAllocationGating {
    /// @notice Constructor
    /// @param _owner The address of the owner of this contract.
    constructor(address _owner, address _allo) {
        // todo:
    }

    /// @notice Modifier to check if the caller is the Allo contract
    modifier onlyAllo() override {
        if (msg.sender != address(allo) && address(allo) != address(0)) {
            revert UNAUTHORIZED();
        }
        _;
    }

    modifier onlyPoolManager() override(BaseStrategy) {
        // todo:
        _;
    }

    modifier onlyPoolManagerWithPoolId(uint256 _poolId) override {
        // todo:
        _;
    }

    receive() external payable {}
}
```

codenamejason: Was experimenting with some implementations...

codenamejason: Going to revert this change.

contracts/core/Allo.sol
```diff
@@ -379,7 +379,7 @@ contract Allo is Transfer, Initializable, Ownable, AccessControl {
     /// @notice allocate to multiple pools
     /// @param _poolIds ids of the pools
     /// @param _datas encoded data unique to the allocation strategy for that pool
-    function batchAllocate(uint256[] calldata _poolIds, bytes[] memory _datas) external payable {
+    function batchAllocate(uint256[] calldata _poolIds, bytes[] memory _datas) external {
```
KurtMerbeth, line 382: I removed payable from `batchAllocate`, because when the `batchAllocate` function is called, it iterates over the `_allocate` function for each pool and sends the entire msg.value with each allocation. However, this approach does not work as intended.

I currently see two ways to deal with it:
- Remove the payable keyword and use batchAllocate only for erc20 token (done)
- Modify the code to allocate the appropriate amount of msg.value for each allocation call

@thelostone-mc @codenamejason @zobront

contracts/core/Allo.sol
KurtMerbeth: makes sense
fixed in [690d316](https://github.com/allo-protocol/allo-v2/pull/36/commits/690d316029565766e77a4e464b493337dbfe7664)

contracts/strategies/BaseStrategy.sol
```diff
@@ -36,15 +36,15 @@ abstract contract BaseStrategy is IStrategy {
...
+    modifier onlyAllo() virtual {
         if (msg.sender != address(allo) && address(allo) != address(0)) {
             revert UNAUTHORIZED();
         }
         _;
     }
 
     /// @notice Modifier to check if the caller is a pool manager
-    modifier onlyPoolManager() {
+    modifier onlyPoolManager() virtual {
```
thelostone-mc, line 47: why would we want to make this virtual ?

contracts/strategies/allocation/alt-registry/AltRegistryGating.sol
```diff
@@ -68,7 +68,7 @@ abstract contract AltRegistryGating is BaseAllocationStrategy {
     function registerRecipients(bytes memory _data, address) external payable override returns (address) {
         (address project, address payoutAddress) = abi.decode(_data, (address, address));
         if (project == address(0) || payoutAddress == address(0)) {
-            revert INVALID_ADDRESS();
+            revert BaseStrategy_INVALID_ADDRESS();
```
thelostone-mc, line 71: ^ Should we just keep them consistent and avoid declaring the error twice ?

contracts/core/Allo.sol
thelostone-mc: > Remove the payable keyword and use batchAllocate only for erc20 token

Ah this seems to be the cleanest way to go about this. 
This problem would persist even if we introduced OZ multi-call right ?

contracts/strategies/allocation/BaseAllocationStrategy.sol
thelostone-mc: It was intended to keep things consistent.

All common logic across strategies lives in BaseStrategy
All common logic across allocation strategies lives in BaseAllocationStrategy
All common logic across distribution strategies lives in BaseDistributionStrategy

KurtMerbeth: I see the confusion, since the `BaseAllocationStrategy` has basically no "unique logic", but on the other hand it combines the `BaseStrategy` and the `IAllocationStrategy`

So instead of importing the `BaseStrategy` and `IAllocationStrategy`, devs only need to import the `BaseAllocationStrategy`.

So no matter which strategy you want to build, the approach is the same. Either import the `BaseAllocationStrategy` or the `BaseDistributionStrategy` and start building.

contracts/core/Registry.sol
```diff
@@ -258,4 +286,12 @@ contract Registry is AccessControl {
...
         return keccak256(abi.encodePacked(_nonce, msg.sender));
     }
+
+    /// @notice Transfer thefunds recovered  to the recipient
+    /// @param _token The address of the token to transfer
+    /// @param _recipient The address of the recipient
+    function recoverFunds(address _token, address _recipient) external onlyIdentityOwner(ALLO_OWNER) {
+        uint256 amount = _token == address(0) ? address(this).balance : ERC20(_token).balanceOf(address(this));
+        _transferAmount(_token, _recipient, amount);
+    }
```
zobront, line 296: I don't think this function will work. `onlyIdentityOwner` looks up `identitiesById[_identityId].owner` but there is no `identitiesById` for `ALLO_OWNER`.

I think what we want to do is `require(isMemberOfIdentity(ALLO_OWNER, msg.sender));` since we granted the role, which makes them a member.

contracts/core/Allo.sol
zobront: I like this idea as a sanity check, good call. I would recommend renaming the functions to just the variable names `_allocationStrategy.identityId()` so devs can just set the variables as public instead of adding functions.

contracts/strategies/BaseStrategy.sol
```diff
@@ -0,0 +1,118 @@
...
+    /// === Storage Variables ====
+    /// ==========================
+
+    bytes32 internal STRATEGY_IDENTIFIER;
+
+    bytes32 internal identityId;
+    Allo internal allo;
+
+    uint256 internal poolId;
+    bool public initialized;
```
zobront, line 32: Do we need `initialized` or can we just use the `allo` address to check this, given that `allo` cannot be set to 0?

contracts/strategies/BaseStrategy.sol
```diff
@@ -0,0 +1,118 @@
...
+
+    /// ====================================
+    /// =========== Modifier ===============
+    /// ====================================
+
+    /// @notice Modifier to check if the caller is the Allo contract
+    modifier onlyAllo() {
+        if (msg.sender != address(allo) && address(allo) != address(0)) {
```
zobront, line 40: I don't think the second half of this check is necessary. If `msg.sender == address(allo)` then it can't be address(0).

contracts/core/Registry.sol
KurtMerbeth: oh true! good point

PR #37, opened 2023-06-29
Update distribution

thelostone-mc:


contracts/strategies/distribution/SplitterDistributionStrategy.sol
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    modifier isAllo() {
+        if(msg.sender == address(allo)) {
+            revert CALLER_NOT_ALLO();
+        }
+        _;
+    }
+
+    constructor(uint256 _poolId, address _allo) {
```
KurtMerbeth, line 30: `constructor` -> `initialize`

PR #41, opened 2023-07-05
add alt-registry allocation spec

nfrgosselin:


contracts/strategies/allocation/alt-registry/README.md
```diff
@@ -0,0 +1,47 @@
...
+- **Voter eligibility**: pool admin
+- **Voting method**: simple yes/no
+- **Calculation method**: simple yes/no
+- **Allocation shape**: discrete amount
+
+## Spec
+### Custom Variables
+This strategy will need the following custom variables:
+- `poolOpen` - boolean value that represents whether the pool is accepting new applications
+- `alternateRegistry` - location of alternate registry of eligible recipients
```
codenamejason, line 20: I think we should dial in on one specific example like Hypercerts, Passport(when onchain is ready), Givith or Optimism.

EAS is another option which I am very fond of experimenting with.

thelostone-mc (changes requested):
Are we talking about an alternate registry which implements the registry interface / just having a gatekeeper for new applications ?
Additionally what is voting mechanism here ? Is it similar to QF ?

Each of these gatekeepers would be it's own contract, we could define a base contract and implement flavours of them but for starters -> could we pick one usecase and update the spec to reflect that

contracts/strategies/allocation/alt-registry/README.md
0xZakk: Based on how the protocol has developed, will it still need these custom variables? I don't think so

0xZakk:
Put together PR #44, which can be merged into this PR

I think the goal here is just to showcase that you can write a strategy that uses another registry. Similarly, you can write a strategy that doesn't use a registry at all.

PR #44 implements a very simple example of what an alternative registry might look like, which should hopefully help us think through what this strategy will need to do.

That's all we're trying to do here, so the voting method for this allocation strategy can be something really simple (maybe just most votes wins, or something)

nfrgosselin:
> Are we talking about an alternate registry which implements the registry interface / just having a gatekeeper for new applications ? Additionally what is voting mechanism here ? Is it similar to QF ?
> 
> Each of these gatekeepers would be it's own contract, we could define a base contract and implement flavours of them but for starters -> could we pick one usecase and update the spec to reflect that

@thelostone-mc are these questions still relevant given Zakk's mock and our discussion? If not, can you approve and merge

contracts/strategies/allocation/alt-registry/README.md
nfrgosselin: @0xZakk `poolOpen` might not be relevant, but we need somewhere to account for how the strategy will check for eligibility, right?

contracts/strategies/allocation/alt-registry/README.md
```diff
@@ -0,0 +1,47 @@
...
+- `poolOpen`
+- `alternateRegistry`
+
+#### `applytoPool()`
+Potential recipients can apply to the pool via `applyToPool`. When the application is submitted, the strategy uses the following decision tree to determine eligibility:
+
+- Is `poolOpen` true?
+    - If yes, proceed
+    - If no, revert with message that pool is not accepting proposals
+- Is recipient present in `alternateRegistry`?
```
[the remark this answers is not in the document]
nfrgosselin: @codenamejason all I'm trying to articulate here is that the logic for this specific strategy should just be whether or not the address is in a registry. Would you want me write that differently in the spec? or is that something you want to solve for in the code?

thelostone-mc:
Revering the commit to HEAD~1 and pushing the alt registry to the strategy PR

thelostone-mc (approved):

PR #42, opened 2023-07-05
Add no application spec

nfrgosselin:


contracts/strategies/allocation/no-application/README.md
```diff
@@ -0,0 +1,38 @@
+Spec: No application with NFT contract
+---------------------------------
+
+## Overview 
+Some builders have flagged that they want the ability create pools that don't require applications, where anyone who has taken some action is automatically an eligible recipient in their pool. This is a strategy stub to test that format.
```
codenamejason, line 5: Should we pick one of these "actions" and show an example of it?

contracts/strategies/allocation/no-application/README.md
```diff
@@ -0,0 +1,38 @@
...
+- `alternateRegistry` - location of alternate registry of eligible recipients
+
+### Standard Functions
+All standard functions are functions that the given user can call from the `Allo.sol` contract.
+#### `createPool()`
+The identity admin creates a new pool via `createPool`. At this time, the admin can set the following custom variables:
+- `alternateRegistry`
+    - Open question: should the identity admin also specify the logic by which the alternate projects should be accepted?
+
+When the pool is created, all recipients in the `alternateRegistry` (using the specified logic) should be added to the pool with the `Accepted` status.
```
codenamejason, line 28: Why use a status at all in this scenario?

nfrgosselin: That's a good question — there's a baked-in assumption that every recipient-pool pair on the protocol should have a status. We've typically thought about them as "application" statuses, but we're essentially trying to show what a given recipient's relationship to the pool is. 

- Null (no relationship)
- Pending (recipient has asked, admin has not answered yet)
- Rejected (recipient has asked, admin has said no)
- Accepted (recipient has asked, admin has said yes — recipient is in the pool)

contracts/strategies/allocation/no-application/README.md
```diff
@@ -0,0 +1,38 @@
...
+- **Recipient approval**: programmatic
+- **Voter eligibility**: pool admin
+- **Voting method**: simple yes/no
+- **Calculation method**: simple yes/no
+- **Allocation shape**: discrete amount
+
+## Spec
+### Custom Variables
+This strategy will need the following custom variables:
+- `alternateRegistry` - location of alternate registry of eligible recipients
```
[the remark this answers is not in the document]
nfrgosselin: This kinda relates to your other question — I think there are a few different ways someone might want to do this, depending on what they're trying to reference:

- ERC-20: their wallet holds a certain threshold of a given ERC-20
- ERC-721: their wallet holds a specific NFT
- ERC-1155 (i.e. hypercerts): a given wallet has taken a prescribed action or holds some token through an ERC-1155 contract
- Generic: given a contract address, a given wallet address has been verified to have taken a specific action

I guess the question to the dev team is: for the purposes of stress-testing this, how do you want to approach it? Do you want to do one contract for each of these? Do we want to create a generic strategy that does all of it?

contracts/strategies/allocation/no-application/README.md
owocki: there is some interesting discussion about this use case over at https://twitter.com/owocki/status/1675555641254096896

contracts/strategies/allocation/no-application/README.md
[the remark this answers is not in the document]
KurtMerbeth: For the sake of simplicity i would say: we stick to the token examples (ERC-20/721/1155)
The `Generic contract` case will need some more research to find the ideal solution and it feels like it is nothing we should do right now.

contracts/strategies/allocation/no-application/README.md
KurtMerbeth: > they want the ability create pools that don't require applications

in this case nobody will ask and no admin will answer, correct?

for pools like this one the term `ApplicationStatus` is misleading, but as far as i see it:

`Null` unused
`Pending` unsused
only `Rejected` and `Accepted` are returned (does an address matches the criteria or not. eg: `balanceOf(address) >= treshold ? Accepted : Rejected`)

The bigger problem i see right now is:

````
    /// @notice Returns the status of the application
    /// @param applicationId The applicationId of the application
    function getApplicationStatus(uint256 applicationId) external view override returns (ApplicationStatus) {
        return applicationStatus[applicationId];
    }
````

`getApplicationStatus` (and other functions like `getPayout`) require an integer `applicationId` as parameter which is maybe not given in a case where no application is required. (Like discussed in our call yesterday)

Maybe we can use `bytes32` instead to cover all cases. For most strategies it doesn't matter if the `applicationId` is from type `uint256` or `bytes32`.
and in case of an `address` we could simply hash the address as identifier. 
what do you think @0xZakk ?

0xZakk (approved):

0xZakk:
@nfrgosselin - this one looks good to me! I think this one is ready to be merged

PR #53, opened 2023-07-12
Module Architecture & Example Strategy

zobront:
This PR sets the stage for the plan we discussed on the call this morning.

It divides up strategies into five modules:
1) Recipient Eligibility
2) Voter Eligibility
3) Voting & Calculation
4) Allocation
5) Distribution

Each has a very minimal interface to implement, and any strategy can be constructed by combining them.

I've also put together a first full strategy (`strategies/examples/WinnerTakesAllLinearVotes.sol`) by combining five simple modules I put together. It's not a good strategy, just an example of how it all fits together.

Notes are included in all module interfaces for @nfrgosselin to think about them as he puts together specs.

Still pending before merging this is updating Allo.sol to reflect these changes. I believe @thelostone-mc will handle this tomorrow AM. Then we can merge this and start building modules & strategies within it.


codenamejason:
I wasn't sure if I was to work of this PR or not. I made a new one using this PR's code with some updates 🤦🏼

contracts/interfaces/New.sol
```diff
@@ -0,0 +1,133 @@
...
+}
+
+interface IVoterEligibilityModule {
+    function initializeVoterEligibilityModule(bytes _data) external;
+    function isValidVoter(address _voter) public view returns (bool); // maybe should be renamed allocator, unless we change to vote below?
+}
+
+interface IVotingModule {
+    function initializeVotingModule(bytes _data) external;
+    function allocate(bytes memory _data, address _sender) external payable; //  is this the right name given module changes?
```
nfrgosselin, line 64: I was thinking the same thing @zobront — now that we've gone through this exercise I think we can change this back to `vote()`

contracts/interfaces/New.sol
```diff
@@ -0,0 +1,133 @@
...
+    enum RecipientStatus { None, Pending, Accepted, Rejected }
+    function initializeRecipientEligibilityModule(bytes _data) external;
+    function registerRecipients(bytes memory _data, address _sender) external payable returns (address);
+    function getRecipientStatus(address _recipientId) public view returns (RecipientStatus);
+    function allRecipients() public view returns (address[] memory);
+}
+
+interface IVoterEligibilityModule {
+    function initializeVoterEligibilityModule(bytes _data) external;
+    function isValidVoter(address _voter) public view returns (bool); // maybe should be renamed allocator, unless we change to vote below?
```
nfrgosselin, line 59: Agree with keeping this as `isValidVoter()`

contracts/interfaces/New.sol
codenamejason: I think so too.

contracts/strategies/modules/distribution/OpenERC20Distribution.sol
nfrgosselin, line 1: @zobront is this meant to be the proportional ERC20 distribution? (i.e. what we use in QF)

contracts/strategies/modules/voting/IVotingModule.sol
```diff
@@ -0,0 +1,16 @@
...
+pragma solidity ^0.8.0;
+
+import { IStrategy }  from "../../IStrategy.sol";
+
+interface IVotingModule is IStrategy {
+    function initializeVotingModule(bytes _data) external;
+    // initialize the voting itself, which could contain things like an ERC20 used to determine vote count
+
+    function allocate(bytes memory _data, address _sender) external payable;
+    // called via allo.sol by users to allocate votes to a recipient
```
nfrgosselin, line 11: per note above, would recommend changing this to `vote()`

contracts/strategies/modules/voting/IVotingModule.sol
```diff
@@ -0,0 +1,19 @@
...
+
+interface IVotingModule is IStrategy {
+    function initializeVotingModule(bytes _data) external;
+    // initialize the voting itself, which could contain things like an ERC20 used to determine vote count
+
+    function vote(bytes memory _data, address _sender) external payable;
+    // called via allo.sol by users to allocate votes to a recipient
+    // this will update some data in this contract to store votes, etc.
+
+    function getResults(address[] memory _recipientId, bytes memory _data)
```
nfrgosselin, line 14: @zobront is passing the results as bytes too flexible? I'm wondering if we should be requiring voting to always produce a fixed point number.

zobront: @nfrgosselin It passes the results back as `ResultSummary[] memory`, which is address & number.

The extra data is for the input, which would be for example useful if we were needing to validate against a merkle root. It'll usually be empty.

contracts/strategies/modules/distribution/OpenERC20Distribution.sol
zobront: @nfrgosselin Remember, whether it is proportional or not lives on the Allocation. So this is literally just a way to send funds out. I should add a check against the amount claimed per recipient though, will do that now.

contracts/strategies/modules/recipient/IRecipientEligibilityModule.sol
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.19;
+
+import {IStrategy} from "../../IStrategy.sol";
+
+interface IRecipientEligibilityModule {
```
nfrgosselin, line 6: We will still have situations where certain strategies will want to use different statuses. What's the best way to handle it in this architecture?

zobront: @nfrgosselin I think that's fine. They can add any logic they want to this. They just need to expose the `getRecipientStatus()` function that returns one of the status from our enum. 

They can track internally however they want, and expose additional functions with more options, etc.

The point is that there needs to be one function that simplifies into these four states, as that's what our frontend will be able to read and know will always be consistent.

contracts/strategies/modules/distribution/OpenERC20Distribution.sol
nfrgosselin: Ack, I thought that this was the allocation module. Thanks for clarifying!

contracts/strategies/modules/voting/IVotingModule.sol
nfrgosselin: got it, thanks for clarifying!

contracts/strategies/modules/recipient/IRecipientEligibilityModule.sol
```diff
@@ -0,0 +1,32 @@
...
+    // if there are more steps, additional functions should be added to allow the owner to check
+    // this could also check attestations directly and then Accept
+
+    function getRecipientStatus(address _recipientId) public view returns (RecipientStatus);
+    // simply returns the status of a recipient
+    // probably tracked in a mapping, but will depend on the implementation
+    // for example, the OpenSelfRegistration only maps users to bool, and then assumes Accepted for those
+    // since there is no need for Pending or Rejected
+
+    function allRecipients() public view returns (address[] memory);
```
thelostone-mc, line 29: This might be applicable for every usecase. Consider direct grants where we might have funds being sent to a small subset

contracts/strategies/modules/allocation/WinnerTakeAll.sol
```diff
@@ -0,0 +1,33 @@
+// SPDX-License-Identifier:  AGPL-3.0-only
+pragma solidity 0.8.19;
+
+import {IStrategy} from "../../IStrategy.sol";
+import {IAllocationModule} from "./IAllocationModule.sol";
+
+contract WinnerTakeAll is IAllocationModule, IStrategy {
+    function initializeAllocationModule(bytes memory _data) external {}
+
+    function getPayouts(address[] memory recipientIds, bytes memory _data) public view returns (PayoutSummary[]) {
```
thelostone-mc, line 10: wouldn't this be neater if Allocation + VotingCalculation would be together ?

contracts/strategies/modules/voting/IVotingModule.sol
thelostone-mc, line 14: Still finding it unclear on the benefits of having a voting and allocation module. Could we draw out examples on why this would. be needed  ? 
as opposed to having them merged together

contracts/interfaces/New.sol
```diff
@@ -0,0 +1,135 @@
...
+        for (uint256 i = 0; i < recipientIds.length; i++) {
+            results[i] = votes[recipientIds[i]];
+        }
+        return results;
+    }
+
+    function getPayouts(address[] memory recipientIds) public view returns (PayoutSummary[]) {
+        require(recipientIds.length == 0, "pass empty rec, will use all");
+        address[] memory recipients = allRecipients();
+        uint256[] memory results = getResults(recipients);
```
KurtMerbeth, line 99: I think we shouldn't rely that heavily on arrays.
imagine having an `allReceipients` array with thousands of addresses. it would be very gas consuming and maybe end up in a dos condition by reaching the block gas limit.

I think we need a better way to wire the modules. i could imagine using internal functions as hooks for every state modifying function  between the modules.

```
function allocate(address recipient) external payable {
        votes[recipient] += 1;
    }
 ```
 would become to
 ```
 function allocate(address recipient) external payable {
        votes[recipient] += 1;
       _allocate(address recipient);
    }
 ```
 And the `WinnerTakesAll` module would implement something like:
 
 ```
 address maxAllocationAddress;

function _allocate(address _receipient) internal {
   if(votes[_receipient] > votes[maxAllocationAddress]) {
        maxAllocationAddress = _receipient;
   }
}
// does it makes sense to split it into an internal and a external function instead of a public one?
function getPayouts(address[] memory recipientIds) public view returns (PayoutSummary[]) {
        require(recipientIds.length == 0, "pass empty rec, will use all");

       PayoutSummary[] memory payouts = new PayoutSummary[](1);
        for (uint256 i = 0; i < recipientIds.length; i++) {
            if (recipients[i] == maxAllocationAddress) {
                payouts[0] = PayoutSummary(recipients[i], address(this).balance);
            }
        }
        return payouts;
```

Tbh i'm currently not sure how we would implement the internal hook if multiple modules would need to implement the hook. but it feels like exploring this path could save a lot of heavy calculations and gas.

@zobront @thelostone-mc @codenamejason

contracts/strategies/modules/voter/IVoterEligibilityModule.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.19;
+
+import {IStrategy} from "../../IStrategy.sol";
+
+interface IVoterEligibilityModule {
+    function initializeVoterEligibilityModule(bytes _data) external;
+    // used to initialize voter eligibility, for example with token address if token gated
+
+    function isValidVoter(address _voter) public view returns (bool);
```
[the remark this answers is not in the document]
KurtMerbeth: We just got rid of `vote` a few weeks ago. Please let's not reintroduce it.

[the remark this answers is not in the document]
zobront: @thelostone-mc @KurtMerbeth Agreed!

If we're going to rename voting to allocation, we need a new term for what @nfrgosselin was calling "Allocation Shape". What do we think?

contracts/strategies/modules/allocation/WinnerTakeAll.sol
zobront: Isn't this the conversation we had yesterday? Where @nfrgosselin suggested that they should be separate, to avoid needing to make a "linear vote winner take all", "linear vote proportionate", "linear vote top 3 get paid", "quadratic vote winner take all", "quadratic vote proportionate", etc etc etc.

nfrgosselin (approved):
LGTM

codenamejason:
This looks good!

contracts/strategies/BaseStrategy.sol
```diff
@@ -56,63 +67,31 @@ abstract contract BaseStrategy is IStrategy {
...
+        if (token == _token) {
+            fundedBalance = amount;
+        }
+        uint256 balance = IERC20(_token).balanceOf(address(this));
+        if (balance > fundedBalance) {
+            uint256 excessFunds = balance - fundedBalance;
+            uint256 bounty = (excessFunds * allo.feeSkirtingBounty()) / allo.FEE_DENOMINATOR();
+            excessFunds -= bounty;
+            IERC20(_token).transfer(treasury, excessFunds);
+            IERC20(_token).transfer(msg.sender, bounty);
```
KurtMerbeth, line 94: `msg.sender` will always be Allo
we need to pass the senders address

contracts/strategies/BaseStrategy.sol
```diff
@@ -23,21 +26,29 @@ abstract contract BaseStrategy is IStrategy {
...
+    address public treasury;
 
     /// ====================================
-    /// =========== Modifier ===============
+    /// ========== Constructor =============
+    /// ====================================
+
+    /// @param _allo Address of the Allo contract
+    constructor(address _allo) {
+        allo = Allo(_allo);
```
KurtMerbeth, line 42: setting the allo contract on strategy deployment will make the contract unusable when we have to fully redeploy the Allo contract for some reason. -> we have to redeploy all strategies

On the other hand in this way the implementation is really only useable using Allo. 

Maybe we should discuss the pros and cons of both ways? @thelostone-mc @zobront @codenamejason @nfrgosselin

contracts/strategies/BaseStrategy.sol
KurtMerbeth: And we should introduce a skim event

contracts/strategies/BaseStrategy.sol
```diff
@@ -56,63 +67,31 @@ abstract contract BaseStrategy is IStrategy {
...
-    /// @return uint256 The pool id
-    function getPoolId() external view returns (uint256) {
-        return poolId;
     }
 
-    /// @notice Get the Allo address
-    /// @return address The Allo address
-    function getAllo() external view returns (address) {
-        return address(allo);
-    }
```
KurtMerbeth, line 106: we should keep the getter.
Reason:
Contracts (and Frontends) which interact with the strategies might want to access these variables. Most probably they will only import the Interface contract and not the BaseStrategy contract
Since the variables are defined in the BaseStrategy contract, there is no way to access them using only the Interface.
Yes -> you could query these data also from the Core contract, but it still feels a bit odd to have an interface which does not expose all the main functions and variables.

contracts/core/Allo.sol
```diff
@@ -10,11 +10,27 @@ import "@solady/auth/Ownable.sol";
...
+ *        /::\  \       /:/  /    /:/  /    /::\  \
+ *       /:/\:\  \     /:/  /    /:/  /    /:/\:\  \
+ *      /::\~\:\  \   /:/  /    /:/  /    /:/  \:\  \
+ *     /:/\:\ \:\__\ /:/__/    /:/__/    /:/__/ \:\__\
+ *     \/__\:\/:/  / \:\  \    \:\  \    \:\  \ /:/  /
+ *          \::/  /   \:\  \    \:\  \    \:\  /:/  /
+ *          /:/  /     \:\  \    \:\  \    \:\/:/  /
+ *         /:/  /       \:\__\    \:\__\    \::/  /
+ *         \/__/         \/__/     \/__/     \/__/
+ */
```
KurtMerbeth, line 29: **_Nice!_**
I already planned to introduce something similar :D

contracts/strategies/BaseStrategy.sol
KurtMerbeth: We should also introduce safeTransfer again

contracts/core/Allo.sol
```diff
@@ -285,9 +253,10 @@ contract Allo is Transfer, Initializable, Ownable, AccessControl {
 
         Pool memory pool = Pool({
             identityId: _identityId,
-            allocationStrategy: _allocationStrategy,
-            distributionStrategy: _distributionStrategy,
+            strategy: _strategy,
             metadata: _metadata,
+            token: _token,
+            amount: 0, // this value is updated in _fundPool
```
KurtMerbeth, line 259: we don't need token and amount here, since these values are stored on the strategy
these values SHOULD always be in sync, but we cannot guarantee this, as we also (eventually) allow custom strategies.

i would propose to have a single source of truth approach here and remove it from the struct and fully rely on the strategy.

contracts/core/Allo.sol
```diff
@@ -10,11 +10,27 @@ import "@solady/auth/Ownable.sol";
 
 import {Metadata} from "./libraries/Metadata.sol";
 import {Clone} from "./libraries/Clone.sol";
-import "./libraries/Transfer.sol";
-import "../interfaces/IAllocationStrategy.sol";
-import "../interfaces/IDistributionStrategy.sol";
-import "./Registry.sol";
-
+import {Transfer} from "./libraries/Transfer.sol";
+import {BaseStrategy} from "../strategies/BaseStrategy.sol";
```
KurtMerbeth, line 14: we should move all necessary functions/getter into the interface and not use an abstract contract instead of the interface everywhere

contracts/core/Allo.sol
```diff
@@ -513,24 +427,72 @@ contract Allo is Transfer, Initializable, Ownable, AccessControl {
...
     /// @param _data encoded data unique to the allocation strategy for that pool
+    function allocate(uint256 _poolId, bytes memory _data) external payable {
+        _allocate(_poolId, _data);
+    }
+
+    /// @notice vote to multiple pools
+    /// @param _poolIds ids of the pools
+    /// @param _datas encoded data unique to the allocation strategy for that pool
+    function batchAllocate(uint256[] calldata _poolIds, bytes[] memory _datas) external {
+        for (uint256 i = 0; i < _poolIds.length;) {
```
KurtMerbeth, line 464: we should cache the array length to save some gas
```
uint526 length = _poolIds.length;
for (uint256 i = 0; i < length;) { ... }
```

and we need to add an input validation (`_poolId.length == _data.length`)

contracts/core/Allo.sol
```diff
@@ -513,24 +427,72 @@ contract Allo is Transfer, Initializable, Ownable, AccessControl {
...
+    /// ======= Internal Functions =========
+    /// ====================================
+
+    /// @notice passes _data & msg.sender through to the strategy for that pool
+    /// @param _poolId id of the pool
+    /// @param _data encoded data unique to the strategy for that pool
     function _allocate(uint256 _poolId, bytes memory _data) internal {
-        pools[_poolId].allocationStrategy.allocate{value: msg.value}(_data, msg.sender);
+        pools[_poolId].strategy.allocate{value: msg.value}(_data, msg.sender);
     }
```
KurtMerbeth, line 488: should we check if the pool exists?

contracts/strategies/BaseStrategy.sol
zobront: - Introduced safeTransfer using the Transfer library
- Added Skim event
- Removed `onlyAllo`, this should be called directly, not via Allo

contracts/core/Allo.sol
thelostone-mc: LMAO

contracts/core/Allo.sol
```diff
@@ -29,8 +45,9 @@ contract Allo is Transfer, Initializable, Ownable, AccessControl {
     /// @notice Struct to hold details of an Pool
     struct Pool {
         bytes32 identityId;
-        IAllocationStrategy allocationStrategy;
-        IDistributionStrategy distributionStrategy;
+        BaseStrategy strategy;
```
thelostone-mc: Shouldn't we use the interface instead ? 
(This goes with having getter functions for the storage variable on BaseStrategy)

contracts/core/Allo.sol
```diff
@@ -53,30 +70,32 @@ contract Allo is Transfer, Initializable, Ownable, AccessControl {
...
+    mapping(address => uint256) private _nonces;
+
     /// @notice Pool.id -> Pool
     mapping(uint256 => Pool) public pools;
 
     /// @notice Strategy -> bool
     mapping(address => bool) public approvedStrategies;
 
+    /// @notice Reward for catching fee skirting (1e18 = 100%)
+    uint256 public feeSkirtingBounty;
```
thelostone-mc, line 89: rename this to `feeSkirtingBountyPercentage` ?

contracts/core/Allo.sol
```diff
@@ -151,104 +170,55 @@ contract Allo is Transfer, Initializable, Ownable, AccessControl {
...
     ) external payable returns (uint256 poolId) {
-        address allocationStrategy;
-        address distributionStrategy;
-
-        // DISCUSS: We either
-        // - allow cloning of every contract (use the bool)
-        // - or we force cloning of only approved contracts
-
-        if (_allocationStrategy == address(0) || _distributionStrategy == address(0)) {
+        if (_strategy == address(0)) {
```
thelostone-mc, line 190: need a check here to ensure the _strategy is not an approved one. We don't want to use those

contracts/core/Allo.sol
thelostone-mc: It looks like we want to store it on Pool struct. I quite like that logic. The skirt logic works with this as well.

contracts/core/Allo.sol
thelostone-mc: wouldn't it just fail is pool doesn't exist ?

contracts/core/libraries/Payouts.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.19;
+
+library Payouts {
+    struct PayoutSummary {
+        address recipient;
+        uint256 amount;
+        uint256 percentage;
```
thelostone-mc, line 8: ^ Do we want to standardise this to just use amounts and keep it cleaner. ?

contracts/core/Allo.sol
zobront: If it doesn't exist, strategy will be `address(0)` so it'll revert. We can check if you want custom error message in that case.

contracts/strategies/BaseStrategy.sol
thelostone-mc: What's the downside of doing it the earlier way ? Having it being set on the initialize function where it would be set to msg.sender ?

contracts/strategies/BaseStrategy.sol
zobront: Changed the names to reflect the variable names, made them public variables, and just added getters to the interface to match with this. We can change if you guys prefer the names with `get`.

thelostone-mc: +1 on keeping getter

contracts/strategies/BaseStrategy.sol
```diff
@@ -56,63 +67,31 @@ abstract contract BaseStrategy is IStrategy {
...
-    /// @notice Returns the strategy identifier
-    function getStrategyIdentifier() external view returns (bytes32) {
-        return STRATEGY_IDENTIFIER;
-    }
-
-    /// @notice sets the strategy identifier
-    /// @param _strategyIdentifier the strategy identifier
-    function _setStrategyIdentifier(string memory _strategyIdentifier) internal {
-        STRATEGY_IDENTIFIER = keccak256(abi.encode(_strategyIdentifier));
+    function skim(address _token) external virtual override onlyAllo {
```
thelostone-mc: Thoughts on tweaking the logic to 


```
    /// @notice Can be called when contract balance is greater than the amount captured in Allo.pool
    /// the remaining balance will be split between the treasury and the msg.sender
    /// @param _token Address of the token
    function skim(address _token) external override {
        (,, address token, uint256 amount,,,) = allo.pools(poolId);

        uint256 balanceCapturedInPool = (token == _token) ? amount : 0;
        uint256 balanceInStrategy =
            _token == address(0) ? address(this).balance : IERC20(_token).balanceOf(address(this));

        if (balanceInContract > balanceCapturedInPool) {
            uint256 excessFunds = balanceInStrategy - balanceCapturedInPool;
            uint256 bounty = (excessFunds * allo.feeSkirtingBounty()) / allo.FEE_DENOMINATOR();
            excessFunds -= bounty;
            IERC20(_token).transfer(allo.treasury(), excessFunds);
            IERC20(_token).transfer(msg.sender, bounty);
        }
    }
   ```

**Review by thelostone-mc — COMMENTED** [review]
Left a few comments

File: contracts/strategies/IStrategy.sol [hunk 23]
```diff
@@ -0,0 +1,76 @@
...
+    // if there are more steps, additional functions should be added to allow the owner to check
+    // this could also check attestations directly and then Accept
+
+    function getRecipientStatus(address _recipientId) external view returns (RecipientStatus);
+    // simply returns the status of a recipient
+    // probably tracked in a mapping, but will depend on the implementation
+    // for example, the OpenSelfRegistration only maps users to bool, and then assumes Accepted for those
+    // since there is no need for Pending or Rejected
+
+    function isValidAllocater(address _voter) external view returns (bool);
```
thelostone-mc, line 65: this should. be _allocator

contracts/strategies/IStrategy.sol
```diff
@@ -0,0 +1,76 @@
...
+
+    function allocate(bytes memory _data, address _sender) external payable;
+    // only called via allo.sol by users to allocate votes to a recipient
+    // this will update some data in this contract to store votes, etc.
+
+    function distribute(address[] memory _recipientIds, bytes memory _data, address _sender) external;
+    // this will distribute tokens to recipients
+    // most strategies will track a TOTAL amount per recipient, and a PAID amount, and pay the difference
+    // this contract will need to track the amount paid already, so that it doesn't double pay
+}
```
thelostone-mc, line 76: ^ Doens't it makes sense to have a getPayout / getPayouts to expose to the frontend

contracts/strategies/BaseStrategy.sol
KurtMerbeth: @thelostone
Anyone can clone and use the implementation
That's the only downside I see

contracts/core/libraries/Payouts.sol
zobront: Since we just have `distribute` on the interface now, it's up to individual strategies how they implement this. I'll get rid of it.

contracts/strategies/IStrategy.sol
zobront: Thank you!

contracts/core/Allo.sol
thelostone-mc: i think it's fine @KurtMerbeth 
we could add that in later if we feel strongly about it

contracts/strategies/IStrategy.sol
zobront: Yeah I think there will be a number of view functions for the front end to add to this IStrategy. This is everything that will be used on chain, which will allow us to start building.

contracts/strategies/BaseStrategy.sol
zobront: Done.

contracts/core/Allo.sol
zobront: We should be set for all approved ones to make sure the `initialize()` call fails, but added this check for extra safety.

contracts/core/Allo.sol
zobront: Done.

contracts/core/Allo.sol
zobront: Done.

codenamejason (approved):

thelostone-mc (approved):

PR #58, opened 2023-07-13
Strategy: ERC721 Votes (1 NFT, 1 Vote) with Winner Take All Distribution

zobront:
Here is a rough V1 of a strategy. It allows anyone with a given NFT to vote. After a fixed amount of time, the recipient with the most votes receives the full funding pool.

Note: This strategy has surfaced the fact that, once a distribution is made, Allo.sol needs the ability to update the `fundedAmount` to keep the balance and that value in sync. 

I'll think more about the solution before implementing it and finalizing this startegy. My current thinking is that it should expose a `decreasePoolFunding()` function that only a strategy can call for itself and allows it to lower the amount of funding when distributions are made. Once I've thought more will open a separate PR for that.


contracts/strategies/BaseStrategy.sol
```diff
@@ -56,63 +62,31 @@ abstract contract BaseStrategy is IStrategy {
...
-            revert BaseStrategy_INVALID_ADDRESS();
-        }
-
-        initialized = true;
-
-        _setStrategyIdentifier(_strategyIdentifier);
+    /// @dev This function is called by Allo.sol
+    function initialize(bytes32 _identityId, uint256 _poolId, bytes memory _data) external virtual onlyAllo {
+        require(_identityId != bytes32(0), "invalid identity id");
+        require(identityId == bytes32(0), "already initialized");
```
KurtMerbeth, line 71: we should use error instead of require

contracts/strategies/NFTVoteWinnerTakeAll.sol
```diff
@@ -0,0 +1,90 @@
...
+    address[] memory recipients;
+    mapping(address => bool) public isRecipient;
+    mapping(uint => bool) public isVoted;
+    mapping(address => uint) public votes;
+
+    function initialize(bytes32 _identityId, uint256 _poolId, bytes memory _data) external override {
+        super.initialize(_identityId, _poolId, _data);
+        (nft, startTime, endTime) = ERC721(abi.decode(_data, (address, uint, uint)));
+
+        require(startTime < endTime, "start time must be before end time");
```
KurtMerbeth, line 24: we should stick to error

contracts/strategies/NFTVoteWinnerTakeAll.sol
```diff
@@ -0,0 +1,90 @@
...
+    address winner;
+
+    address[] memory recipients;
+    mapping(address => bool) public isRecipient;
+    mapping(uint => bool) public isVoted;
+    mapping(address => uint) public votes;
+
+    function initialize(bytes32 _identityId, uint256 _poolId, bytes memory _data) external override {
+        super.initialize(_identityId, _poolId, _data);
+        (nft, startTime, endTime) = ERC721(abi.decode(_data, (address, uint, uint)));
```
KurtMerbeth, line 22: i've never seen this syntax with parsing the whole abi.decode. does it actually works? does it also work if two addresses are included?

contracts/strategies/NFTVoteWinnerTakeAll.sol
```diff
@@ -0,0 +1,90 @@
...
+    function registerRecipients(bytes memory _data, address _sender) external payable returns (address) {
+        require(block.timestamp >= startTime, "voting has not started");
+        require(block.timestamp <= endTime, "voting has ended");
+
+        address[] memory newRecipients = abi.decode(_data, (address[]));
+
+        // check to ensure gas for determining winner doesn't exceed block gas limit (just made up number for now)
+        require(recipients.length + newRecipients.length < 1000, "max 1000 recipients");
+
+        for (uint i; i < newRecipients.length; i++) {
```
KurtMerbeth, line 37: we should cache the length and wrap i++ in unchecked block

contracts/strategies/NFTVoteWinnerTakeAll.sol
```diff
@@ -0,0 +1,90 @@
...
+        } else {
+            return RecipientStatus.None;
+        }
+    }
+
+    function isValidAllocator(address _voter) public view returns (bool) {
+        return nft.balanceOf(_voter) > 0;
+    }
+
+    function allocate(bytes memory _data, address _sender) external payable {
```
KurtMerbeth, line 55: function should implement `onlyAllo` modifier

contracts/strategies/NFTVoteWinnerTakeAll.sol
```diff
@@ -0,0 +1,90 @@
...
+        require(block.timestamp >= startTime, "voting has not started");
+        require(block.timestamp <= endTime, "voting has ended");
+
+        (uint256[] memory ids, address recipient) = abi.decode(_data, (uint256[], address));
+        uint numVotes = ids.length;
+        for (uint i; i < numVotes; i++) {
+            require(!isVoted[ids[i]], "already voted");
+            require(nft.ownerOf(ids[i]) == msg.sender, "not owner of NFT");
```
KurtMerbeth, line 63: we should use `_sender` instead of `msg.sender` since the function will only be callable by Allo

contracts/strategies/NFTVoteWinnerTakeAll.sol
```diff
@@ -0,0 +1,90 @@
...
+            isVoted[votes[i]] = true;
+        }
+        votes[recipient] += numVotes;
```
KurtMerbeth, line 66: we can introduce a global `address maxVotes` variable

and add the following logic here:
```
if(votes[receipient] > votes[maxVotes]) {
   maxVotes = receipient;
}
```
This will enable us to remove the max receipients restriction and the expensive operations in `determineWinner`

contracts/strategies/NFTVoteWinnerTakeAll.sol
```diff
@@ -0,0 +1,90 @@
...
+        require(winner == address(0), "winner already determined");
+
+        uint256 maxVotes;
+        address winner;
+        for (uint i; i < recipients.length; i++) {
+            if (votes[i] > maxVotes) {
+                maxVotes = votes[i];
+                winner = i;
+            }
+        }
```
KurtMerbeth, line 80: expensive operation can be removed by changing the allocate function a bit [see the following comment](https://github.com/allo-protocol/allo-v2/pull/58/files#r1263554934)

contracts/strategies/NFTVoteWinnerTakeAll.sol
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    function registerRecipients(bytes memory _data, address _sender) external payable returns (address) {
+        require(block.timestamp >= startTime, "voting has not started");
+        require(block.timestamp <= endTime, "voting has ended");
+
+        address[] memory newRecipients = abi.decode(_data, (address[]));
+
+        // check to ensure gas for determining winner doesn't exceed block gas limit (just made up number for now)
+        require(recipients.length + newRecipients.length < 1000, "max 1000 recipients");
```
KurtMerbeth, line 35: receipients count restriction can be removed 
[see the following comment](https://github.com/allo-protocol/allo-v2/pull/58/files#r1263554934)

contracts/strategies/NFTVoteWinnerTakeAll.sol
```diff
@@ -0,0 +1,90 @@
...
+        for (uint i; i < recipients.length; i++) {
+            if (votes[i] > maxVotes) {
+                maxVotes = votes[i];
+                winner = i;
+            }
+        }
+        return winner;
+    }
+
+    function distribute(address[] memory _recipientIds, bytes memory _data, address _sender) external {
```
KurtMerbeth, line 84: function should be restricted to onlyAllo

contracts/strategies/NFTVoteWinnerTakeAll.sol
zobront: Very smart solution, thanks!

contracts/strategies/NFTVoteWinnerTakeAll.sol
```diff
@@ -0,0 +1,144 @@
...
+
+    function getRecipientStatus(address _recipientId) public view returns (RecipientStatus) {
+        if (isRecipient[_recipientId]) {
+            return RecipientStatus.Accepted;
+        } else {
+            return RecipientStatus.None;
+        }
+    }
+
+    function isValidAllocator(address _voter) public view returns (bool) {
```
thelostone-mc, line 91: ^ _allocator :P

codenamejason (approved):

PR #62, opened 2023-07-14
Foundation fixes

zobront:
`private` variables can't be read by contracts that inherit them. I removed that from BaseStrategy so strategies can access these variables.

Also changed `initialize()` to public from external so strategies can call `super.initialize()`.

### DIFF
--- a/contracts/core/Allo.sol
+++ b/contracts/core/Allo.sol
@@ -110,6 +110,8 @@ contract Allo is Transfer, Initializable, Ownable, AccessControl {
 
     event PoolClosed(uint256 indexed poolId);
 
+    event PoolTotalFundingDecreased(uint256 indexed poolId, uint256 prevAmount, uint256 newAmount);
+
     event TreasuryUpdated(address treasury);
 
     event FeePercentageUpdated(uint256 feePercentage);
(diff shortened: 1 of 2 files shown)


thelostone-mc (approved):

contracts/strategies/BaseStrategy.sol
```diff
@@ -12,9 +12,9 @@ abstract contract BaseStrategy is IStrategy, Transfer {
     /// === Storage Variables ====
     /// ==========================
 
-    Allo private immutable allo;
-    uint256 private poolId;
-    string private strategyName;
+    Allo immutable allo;
+    uint256 poolId;
+    string strategyName;
```
thelostone-mc, line 17: ^ we should make them internal

contracts/core/Allo.sol
```diff
@@ -426,6 +428,17 @@ contract Allo is Transfer, Initializable, Ownable, AccessControl {
         _transferAmount(_token, _recipient, amount);
     }
 
+    function decreasePoolTotalFunding(uint256 _poolId, uint256 _amountToDecrease) external {
```
thelostone-mc, line 431: Ah this is to signal a reduction anytime a payout happens ?

contracts/strategies/BaseStrategy.sol
zobront: Internal is default, but added in commit below for clarity.

thelostone-mc (approved):

PR #64, opened 2023-07-14
chore: rewrite rfp

### LINKED ISSUE #77: Update simple rfp strategy
Spec: #67 

This is part of the work to create our initial batch of allocation strategies. This spec is to update the simple [RFP strategy here](https://github.com/allo-protocol/allo-v2/pull/64).

[Strategy hitlist](https://docs.google.com/spreadsheets/d/1hEFSOPe55GNJN8skdAzpHW-zCA2K0sWMd3kyHC0TeO0/edit#gid=0)
[Allocation module overview](https://docs.google.com/document/d/1qoOP07oMKzUCyfb4HbnyeD6ZYEQa004i5Zwqoy7-Ox8/edit#heading=h.hu89fnhk88ap)
[Original strategy brainstorm](https://docs.google.com/document/d/1QBOAmYhwgxp-e9uuuOaPdCT674fLsn2vLH74-I8PqkQ/edit)

### LINKED ISSUE #78: Write RFP Committee strategy
Spec: #68  

[Strategy hitlist](https://docs.google.com/spreadsheets/d/1hEFSOPe55GNJN8skdAzpHW-zCA2K0sWMd3kyHC0TeO0/edit#gid=0)
[Allocation module overview](https://docs.google.com/document/d/1qoOP07oMKzUCyfb4HbnyeD6ZYEQa004i5Zwqoy7-Ox8/edit#heading=h.hu89fnhk88ap)
[Original strategy brainstorm](https://docs.google.com/document/d/1QBOAmYhwgxp-e9uuuOaPdCT674fLsn2vLH74-I8PqkQ/edit)

thelostone-mc:
- Implement rfp-simple
- Implement rfp-committee

Fixes #77 
Fixes #78


contracts/core/libraries/Payouts.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.19;
+
+contract Payouts {
+    struct PayoutSummary {
+        address recipient;
+        uint256 amount;
+        uint256 percentage;
+    }
+}
```
zobront, line 10: Should this be in something like `strategies/libraries/`? I'm not sure, just seems like it doesn't belong in `core/` since it's not used anywhere in there.

contracts/strategies/RFPStrategy.sol
```diff
@@ -0,0 +1,169 @@
...
+
+    constructor(address _allo, string memory _name) BaseStrategy(_allo, _name) {}
+
+    /// ====================================
+    /// =========== Functions ==============
+    /// ====================================
+
+    function initialize(uint256 _poolId, bytes memory _data) public override {
+        super.initialize(_poolId, _data);
+    }
```
zobront, line 64: I think you can skip this — if it's just calling `super.initialize()` then not including the function will do the same thing (default to the function in the inherited contract).

contracts/strategies/RFPStrategy.sol
```diff
@@ -0,0 +1,169 @@
...
+            abi.decode(_data, (address, uint256, Metadata));
+
+        Recipient memory recipient = Recipient({
+            payoutAddress: payoutAddress,
+            proposedAmount: proposedAmount,
+            recipientStatus: RecipientStatus.Pending
+        });
+
+        address recipientId = address(uint160(_counter++));
+        recipients[recipientId] = recipient;
```
zobront, line 84: Is the reason to make this a counter so that the same address can submit twice? I thought we settled on recipientId being the address of the recipient, but this does seem to be a problem with that format.

contracts/strategies/RFPStrategy.sol
```diff
@@ -0,0 +1,169 @@
...
+                recipients[recipientId].recipientStatus = RecipientStatus.Accepted;
+                acceptedRecipientId = recipientId;
+            }
+
+            recipients[recipientId].recipientStatus = RecipientStatus.Rejected;
+
+            unchecked {
+                i++;
+            }
+        }
```
zobront, line 128: Do we need to update to rejected? It seems worthwhile, but also introduces a loop that could get very long, and no way to get around it to just set the winner to Approved, which is all that's really needed for the rest of the contract to function. 

The better solution I think is to have `getRecipientStatus()` perform this logic:
- if (payoutReady) 
    - if Accepted => Accepted
    - if Pending => Rejected
- else: recipientStatus

zobront (commented):
This looks great. Left a few comments but otherwise looks ready to merge.

contracts/core/libraries/Payouts.sol
codenamejason: I agree, thinking about where to put it 🤔 and others that may evolve like it.

codenamejason (dismissed):

contracts/strategies/RFPStrategy.sol
```diff
@@ -0,0 +1,171 @@
...
+
+import {BaseStrategy} from "./BaseStrategy.sol";
+import {Metadata} from "../core/libraries/Metadata.sol";
+import {ReentrancyGuard} from "@openzeppelin/security/ReentrancyGuard.sol";
+
+contract RFPStrategy is BaseStrategy, ReentrancyGuard {
+    /// @notice Struct to hold details of an recipient
+    struct Recipient {
+        address payoutAddress;
+        uint256 proposedAmount;
```
0xZakk, line 12: Do we need the proposed amount? Wouldn't the entire pool go to the one recipient?

KurtMerbeth: Yes, it's part of the spec.
It's named different in the spec, but here is the link: https://github.com/allo-protocol/allo-v2/blob/dd1b6b137b44f780717b4228d107e384b90fa2ad/contracts/strategies/allocation/rfp-simple/README.md?plain=1#L23

contracts/core/libraries/Payouts.sol
thelostone-mc: removing this for now. we'll handle this when we create libraries/module

contracts/strategies/RFPStrategy.sol
```diff
@@ -0,0 +1,239 @@
...
+    /// ===============================
+
+    /// @notice Checks if msg.sender is eligible for RFP allocation
+    /// @param _recipientId Id of the recipient
+    function getRecipientStatus(address _recipientId) public view override returns (RecipientStatus status) {
+        return recipients[_recipientId].recipientStatus;
+    }
+
+    /// @notice Get the payout summary for accepted recipient
+    function getPayouts(address[] memory, bytes memory, address) external view returns (PayoutSummary[] memory) {
```
thelostone-mc, line 88: What should this return ?

contracts/strategies/RFPStrategy.sol
```diff
@@ -0,0 +1,239 @@
...
+    /// @notice Struct to hold details of an recipient
+    struct Recipient {
+        address payoutAddress;
+        uint256 proposalBid;
+        RecipientStatus recipientStatus;
+    }
+
+    /// @notice Struct to hold milestone details
+    struct Milestone {
+        uint256 amountPercentage;
```
thelostone-mc, line 23: This can be set on init. But  should we also track: actual amount ?

contracts/strategies/RFPStrategy.sol
```diff
@@ -0,0 +1,278 @@
...
+    /// ================================
+    /// ========== Storage =============
+    /// ================================
+
+    /// @notice Struct to hold details of an recipient
+    struct Recipient {
+        address recipientAddress;
+        uint256 proposalBid;
+        RecipientStatus recipientStatus;
+    }
```
nfrgosselin, line 20: Don't we need a recipientID and a payout address here?

KurtMerbeth: recipientId is not required, since the relation is stored in
```solidity
 /// @notice recipientId -> Recipient
 mapping(address => Recipient) public recipients;
```


The payout address is the same as `recipientAddress`

contracts/strategies/RFPStrategy.sol
```diff
@@ -0,0 +1,278 @@
...
+        return milestones[_milestoneId].milestoneStatus;
+    }
+
+    /// ===============================
+    /// ======= External/Custom =======
+    /// ===============================
+
+    /// @notice Set milestones for RFP pool
+    /// @param _milestones The milestones to be set
+    function setMilestones(Milestone[] memory _milestones) external onlyPoolManager(msg.sender) {
```
KurtMerbeth, line 118: function allows overwriting old and paid milestones
I'm not sure if it's a problem. just leave this note here.

wdyt about:
```solidity
        uint256 paidMilestones = upcomingMilestone == 0 ? 0 : upcomingMilestone - 1;
        uint256 milestonesLength = _milestones.length;
        if(milestoneLength <= paidMilestones) {
             revert INVALID_MILESTONE();
        } else {
             milestoneLength -= paidMilestones;
        }
        for (uint256 i = paidMilestones; i < milestonesLength;) {
            milestones.push(_milestones[i]);

            unchecked {
                i++;
            }
        }
```

contracts/strategies/RFPStrategy.sol
```diff
@@ -0,0 +1,278 @@
...
+    function submitMilestone(Metadata calldata _metadata) external {
+        if (acceptedRecipientId != msg.sender && !_isIdentityManager(acceptedRecipientId, msg.sender)) {
+            revert UNAUTHORIZED();
+        }
+
+        if (upcomingMilestone > milestones.length) {
+            revert INVALID_MILESTONE();
+        }
+
+        milestones[upcomingMilestone].metadata = _metadata;
```
KurtMerbeth, line 142:
```solidity
        milestones[upcomingMilestone].metadata = _metadata;
        milestones[upcomingMilestone].milestoneStatus = RecipientStatus.Pending;
```

should become
```solidity
Milestone storage  ms = milestones[upcomingMilestone];
ms.metadata = _metadata;
ms.milestoneStatus = RecipientStatus.Pending;
```

contracts/strategies/RFPStrategy.sol
```diff
@@ -0,0 +1,278 @@
...
+    /// @param _sender The sender of the allocation
+    function _allocate(bytes memory _data, address _sender) internal override nonReentrant onlyPoolManager(_sender) {
+        if (acceptedRecipientId != address(0)) {
+            revert RECIPIENT_ALREADY_ACCEPTED();
+        }
+
+        acceptedRecipientId = abi.decode(_data, (address));
+
+        uint256 _recipientsCounter = _recipientIds.length;
+        for (uint256 i = 0; i < _recipientsCounter;) {
```
KurtMerbeth, line 226: this for loop can end up in a DOS condition, if the `_recipientsCounter` reaches a certain limit

i would remove the entire for loop and change `getRecipientStatus` to
```solidity
function getRecipientStatus(address _recipientId) public view override returns (RecipientStatus status) {
    Recipient memory recipient = recipients[_recipientId];
    
    if (acceptedRecipientId == address(0) || acceptedRecipientId == _recipientId) {
        return recipient.recipientStatus;
    }
    
    return recipient.recipientStatus > RecipientStatus.None ? RecipientStatus.Rejected : RecipientStatus.None;
}
```

also i would make the `recipients`-mapping private and introduce a getter which replaces the status based on the logic above.

this enables us to remove the for loop entirely

contracts/strategies/RFPStrategy.sol
thelostone-mc: Done

contracts/strategies/RFPStrategy.sol
thelostone-mc: Updated this to be msg.sender / identityId from registry

contracts/strategies/RFPStrategy.sol
thelostone-mc: Ah i see what you're saying but here is what i think what we should update it to 
```
        if (upcomingMilestone != 0) {
            revert MILESTONES_ALREADY_SET();
        }
```

Are you thinking we should allow partial setting of milestones ?

contracts/strategies/RFPStrategy.sol
```diff
@@ -0,0 +1,278 @@
...
+
+        if (upcomingMilestone > milestones.length) {
+            revert INVALID_MILESTONE();
+        }
+
+        IAllo.Pool memory pool = allo.getPool(poolId);
+        Milestone storage milestone = milestones[upcomingMilestone];
+        Recipient memory recipient = recipients[acceptedRecipientId];
+
+        uint256 amount = recipient.proposalBid * milestone.amountPercentage / 1e18;
```
KurtMerbeth, line 262: we need to track the total percentage of all milestones somewhere. Else we would be able to distribute more than 100% of the `proposalBid` in total.
(for example if you have 5 milestones with 25% each)

thelostone-mc: Maybe we loop all milestones when setMilestones is invoked and we ensure total is 100% ?

contracts/strategies/RFPStrategy.sol
```diff
@@ -0,0 +1,278 @@
...
+            revert RECIPIENT_ALREADY_ACCEPTED();
+        }
+
+        address recipientAddress;
+        uint256 proposalBid;
+        Metadata memory metadata;
+
+        // decode data custom to this strategy
+        if (registryGating) {
+            (recipientId, recipientAddress, proposalBid, metadata) =
```
KurtMerbeth, line 190: we cant use the anchor as identyId
since the anchor can change during the pool period

if the anchor changes during the pool period, this check on `submitMilestone` will fail.

to still be able to rely on the anchor we could maybe remove `anchorToIdentityId[identity.anchor] = bytes32(0);` on the `Registry` in `updateIdentityName`
-> we should carefully think about pros and cons

thelostone-mc: @nfrgosselin @zobront @codenamejason ^

contracts/strategies/RFPStrategy.sol
```diff
@@ -0,0 +1,278 @@
...
+
+        emit Distributed(acceptedRecipientId, recipient.recipientAddress, amount, _sender);
+    }
+
+    function _isIdentityManager(address _anchor, address _sender) internal view returns (bool) {
+        IRegistry registry = allo.getRegistry();
+        IRegistry.Identity memory identity = registry.getIdentityByAnchor(_anchor);
+        return registry.isOwnerOrMemberOfIdentity(identity.id, _sender);
+    }
+}
```
KurtMerbeth, line 283: Missing functionalities:
When does the pool end?

WIthdrawing funds

thelostone-mc: Do we need a pool ended variable  ? @nfrgosselin

KurtMerbeth: as a pool owner you might want to close the pool if you can't find a recipient which fits your needs.

also we should think about introducing functions on the interface for:

canRegister(address)
canAllocate(address)
.. or sth similar. It differs from isValidAllocator, since canAllocate could check if:
address == isValidAllocator && acceptedRecipientId == address (0) && ...
=> basically a helper function for frontend implementations

we talked earlier in one of our discussions about it.

nfrgosselin: Confirmed from standup: we want to have a variable in the core Allo interface that indicates whether or not a pool is active

contracts/strategies/RFPStrategy.sol
zobront: I think it is a problem. Let's picture this situation:
- my identity id = 1 and name = Zach Obront => this outputs Anchor 1
- i change my name to Elon Musk => this outputs Anchor 2
- any attestation to Anchor 1 is still considered valid, which means I can submit to projects as Elon Musk (because the anchor is pointing to the new struct that has been updated)

***

I believe the current setup in this contract also limits that at most one recipient can be registered per `recipientId` / `anchor`, which may limit a situation when there are multiple members from one identity that want to submit.

It seems like the ideal solution is that, rather than tracking recipients by `recipientId`, we track them by `recipientAddress`. That means `recipientId` is only a blocker when submitting, to make sure the identity is valid, and after that it's just there for name/metadata. 

***

This still leaves the problem that the link would break if they changed the name (because that would also change the anchor). But we probably want the behavior, not to just assume the new name is valid? 

The only solution to that I can see is that you can never change a name in the Registry. It has to continue to exist forever. You can simply create a new identity in the Registry, with a new name, and then base things on that.

I actually think that may be better, as it ensures historical ones are always there, and doesn't seem to harm users at all.

Thoughts?

nfrgosselin: > The only solution to that I can see is that you can never change a name in the Registry. It has to continue to exist forever. You can simply create a new identity in the Registry, with a new name, and then base things on that.
I actually think that may be better, as it ensures historical ones are always there, and doesn't seem to harm users at all.

This is actually where I thought we ended up. I'm okay with this, as we can normalize an attestation pattern that allows old identities to validate their new identities

codenamejason (approved):
To me this reads a lot better. Nice job!

KurtMerbeth (approved):

PR #72, opened 2023-07-14
add qv gov spec

nfrgosselin:


contracts/strategies/allocation/quadratic-voting-governance/README.md
```diff
@@ -0,0 +1,42 @@
...
+        - For example, if the threshold is set to `2`, then two separate admins must approve or reject an application before its status updates. As soon as the recipient receives either two rejections or two approvals, admins can no longer take any action on that application. 
+
+### Voter eligibility logic
+Pool admins are able to designate a governance token that they want to use as their source of truth. The strategy takes a snapshot of token holdings on an admin-specified date and uses that as its source of truth for the voting round. Voters must hold tokens in that snapshot in order to be eligible. 
+
+### Voting logic
+In this strategy, only the eligible wallets are able to vote. Votes can only be cast during a voting window that is set by the pool admin. Each voter gets a budget of "voice credits", which they can use to "purchase" votes on as many recipients as they would like. The number of votes each voter can give to a recipient is equal to the square root of the number of voice credits they spend on that specific recipient. Votes can be fractional.
```
codenamejason, line 25: This strategy uses a token, is this token created for this strategy or they just issue the `voters` some tokens the admins have? Trying to think of how these "credits" are created and distributed (the budget).

contracts/strategies/allocation/quadratic-voting-governance/README.md
```diff
@@ -0,0 +1,42 @@
...
+            - if pool admin doesn't require Allo identity and the recipient doesn't have one, then `msg.sender` is used as recipient id
+        - pool admin can include optional and required questions in an application, stored in metadata
+            - recipient must answer questions as specified by pool admin
+- `reviewApplication` - function for pool admins to approve or reject a recipient for inclusion in the round.
+    - the pool admins have the ability to set how many admins must confirm a decision in order for an application to be approved or rejected.
+        - For example, if the threshold is set to `2`, then two separate admins must approve or reject an application before its status updates. As soon as the recipient receives either two rejections or two approvals, admins can no longer take any action on that application. 
+
+### Voter eligibility logic
+Pool admins are able to designate a governance token that they want to use as their source of truth. The strategy takes a snapshot of token holdings on an admin-specified date and uses that as its source of truth for the voting round. Voters must hold tokens in that snapshot in order to be eligible.
```
thelostone-mc, line 22: 1. Not sure how we'd take a snapshot on a given day ? We could do something like allow people to cast a vote and then when the voting ends, we recount the votes and check the token balance to see if they are legit votes

codenamejason: I would like clarification on the 'snapshot' technique we will use. Not sure this will be an easy lift out of the gate.

Example of snapshot setup:
https://docs.radiant.so/introduction
https://segment.radiant.so/

codenamejason (approved):

PR #87, opened 2023-07-16
Improvements and fixes on main

KurtMerbeth:


contracts/strategies/IStrategy.sol
```diff
@@ -63,19 +64,18 @@ interface IStrategy {
     // then checking the balance of the contract, and paying the difference
     function skim(address token) external;
 
+    /// @return Input the values you would send to distribute(), get the amounts each recipient in the array would receive
+    function getPayouts(address[] memory _recipientIds, bytes memory _data, address _sender)
```
thelostone-mc, line 68: should this be view ?

KurtMerbeth: yes!
good catch

we also should talk about the return value
i think it should be a struct with an address and an amount. since the receipientId can differ from the address were the funds go to

contracts/strategies/NFTVoteWinnerTakeAll.sol
```diff
@@ -67,17 +65,19 @@ contract NFTVoteWinnerTakeAll is BaseStrategy {
...
-        for (uint256 i; i < newRecipients.length;) {
+        uint256 numRecipientsLength = newRecipients.length;
+        for (uint256 i; i < numRecipientsLength;) {
             isRecipient[newRecipients[i]] = true;
-            recipients.push(newRecipients[i]);
             unchecked {
                 ++i;
             }
         }
+        return address(0);
```
thelostone-mc, line 80: This function at the interface level feels wrong.

`registerRecipients` -> if it registers -> returning 1 address -> doesn't make sense 

would it be an overkill to have registerRecipients meant for pool manager and register meant for user

thelostone-mc (approved):

contracts/core/Allo.sol
```diff
@@ -1,128 +1,70 @@
...
     /// @notice Incremental index
     uint256 private _poolIndex;
 
     /// @notice Allo treasury
-    address payable public treasury;
+    address payable private treasury;
 
     /// @notice Registry of pool creators
-    Registry public registry;
+    IRegistry private registry;
```
0xZakk, line 55: Why make these private?

contracts/core/Allo.sol
```diff
@@ -1,128 +1,70 @@
...
     /// @notice msg.sender -> nonce for cloning strategies
     mapping(address => uint256) private _nonces;
 
     /// @notice Pool.id -> Pool
-    mapping(uint256 => Pool) public pools;
+    mapping(uint256 => Pool) private pools;
 
     /// @notice Strategy -> bool
-    mapping(address => bool) public approvedStrategies;
+    mapping(address => bool) private approvedStrategies;
```
0xZakk, line 64: Are we still using this? Remember we decided to remove the idea of approved strategies - can we remove this?

contracts/strategies/BaseStrategy.sol
```diff
@@ -1,29 +1,29 @@
...
 abstract contract BaseStrategy is IStrategy, Transfer {
     /// ==========================
     /// === Storage Variables ====
     /// ==========================
 
-    Allo internal immutable allo;
+    IAllo internal immutable allo;
     uint256 internal poolId;
-    string internal strategyName;
+    bytes32 internal strategyId;
```
0xZakk, line 17: Why use an ID here instead of a name?

contracts/core/Allo.sol
KurtMerbeth: for now it isfor now it is still an open question
and since it was already implemented, we haven't removed it.

i'm still against removing it :D - but once the final decision is made, we can touch the code again

contracts/strategies/BaseStrategy.sol
KurtMerbeth: bytes32 needs less gas

contracts/core/Allo.sol
KurtMerbeth: We introduced a getter function instead

contracts/core/Allo.sol
```diff
@@ -1,128 +1,70 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 pragma solidity 0.8.19;
 
+import "./IAllo.sol";
+
 import "@openzeppelin-upgradeable/proxy/utils/Initializable.sol";
 import "@openzeppelin-upgradeable/token/ERC20/IERC20Upgradeable.sol";
-import "@openzeppelin-upgradeable/token/ERC20/utils/SafeERC20Upgradeable.sol";
-import {AccessControl} from "@openzeppelin/access/AccessControl.sol";
+import "@openzeppelin/access/AccessControl.sol";
```
codenamejason, line 8: Why use named imports on some and not all? Is there an advantage to this?

contracts/core/Allo.sol
```diff
@@ -1,128 +1,70 @@
...
 
     /// ==========================
     /// === Storage Variables ====
     /// ==========================
 
+    address constant NATIVE = address(0);
     /// @notice Fee percentage
     /// @dev 1e18 = 100%, 1e17 = 10%, 1e16 = 1%, 1e15 = 0.1%
-    uint256 public feePercentage;
+    uint256 private feePercentage;
```
codenamejason, line 43: Did we decide not to make our private variables like this `_feePercentage`? Or are we just doing this for internal functions?

contracts/strategies/BaseStrategy.sol
```diff
@@ -1,29 +1,29 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 pragma solidity 0.8.19;
 
+import "./IStrategy.sol";
+
 import {ERC20} from "@openzeppelin/token/ERC20/ERC20.sol";
```
codenamejason, line 6: We are using both Solady and Openzeppelin for ERC20, 721, etc., should we stick to Solady? Seems we can do everything with that. Not sure about the upgradeable ERC 🤔 I think we still need OZ for that.

codenamejason: Note sure if OZ is using 2612 yet for ERC20 also.

contracts/strategies/BaseStrategy.sol
```diff
@@ -50,16 +50,16 @@ abstract contract BaseStrategy is IStrategy, Transfer {
...
         return allo;
     }
 
     function getPoolId() external view override returns (uint256) {
         return poolId;
     }
 
-    function getStrategyName() external view override returns (string memory) {
-        return strategyName;
+    function getStrategyId() external view override returns (bytes32) {
```
codenamejason, line 61: I like this idea a lot better 🙌

contracts/core/Allo.sol
KurtMerbeth: i think currently only for internal functions, or i miss something :D

contracts/core/Allo.sol
KurtMerbeth: named imports doesn't make sense/there is no big difference in our case.
maybe it's ncer to read for some ppl? idk :D

codenamejason: Just for reading and consistency IMO either way is fine.

thelostone-mc (dismissed):

contracts/core/Allo.sol
```diff
@@ -278,12 +278,37 @@ contract Allo is IAllo, Transfer, Initializable, Ownable, AccessControl {
...
+    /// @param _data encoded data unique to the strategy for that pool
+    function registerRecipient(uint256 _poolId, bytes memory _data) external payable returns (address) {
+        return pools[_poolId].strategy.registerRecipient(_data, msg.sender);
+    }
+
+    /// @notice register to multiple pools
+    /// @notice returns the recipientIds from the strategy
+    /// @param _poolIds id of the pools
+    /// @param _data encoded data unique to strategy for each pool
+    function batchRegisterRecipient(uint256[] memory _poolIds, bytes[] memory _data)
```
codenamejason, line 293: Should we pluralize this? `batchRegisterRecipients()`

codenamejason (approved):
This looks great and accomplishes a lot!!

0xZakk:
LGTM

PR #95, opened 2023-07-19
re-organizing strategies folder

nfrgosselin:


thelostone-mc:
Blocking this until the contract changes get in cause it will create conflicts

PR #115, opened 2023-07-20
Deploy Scripts

codenamejason:


contracts/core/Registry.sol
```diff
@@ -2,11 +2,11 @@
 pragma solidity 0.8.19;
 
 import "./IRegistry.sol";
-import {AccessControl} from "@openzeppelin/access/AccessControl.sol";
+import {AccessControl} from "@openzeppelin/contracts/access/AccessControl.sol";
 import {Metadata} from "./libraries/Metadata.sol";
 import "./libraries/Transfer.sol";
 import "./libraries/Native.sol";
-import {ERC20} from "@solady/tokens/ERC20.sol";
+import {ERC20} from "solady/src/tokens/ERC20.sol";
```
0xZakk, line 9: Hey, I'm not fussed over which format we use, but we should be consistent. In some places we're using the `@` and in others we're not (i.e. here we are for Open Zeppelin but are not for Solady).

.env.sample
```diff
@@ -1,14 +1,24 @@
...
 # Local Node
-ANVIL_PRIVATE_KEY=
\ No newline at end of file
+ANVIL_PRIVATE_KEY=
+
+REPORT_GAS=true
+
+MNEMONIC="test test test test test test test test test test test junk"
+
+OPTIMISTIC_ETHERSCAN_API_KEY=
```
0xZakk, line 18: We're going to need this to scale up to potentially many networks (like 1-4 at launch, then potentially 10 within the year). Can you add comments here that organize the environment variables by network? (i.e. put all the PGN stuff toegether, Fantom stuff together, etc etc)

contracts/core/Registry.sol
0xZakk: I guess I have a slight preference for the `@` and using shorter paths (i.e. mapping `@solady` directly to the `src` directory)

0xZakk: Same with `@openzeppelin` to the `contracts/` directory

yarn-error.log
0xZakk, line 1: Do we need this file? Can this be deleted?

0xZakk (commented):
Overall, I think this is great!

We're going to deploy to 1-4 networks to start but will keep deploying to new chains overtime (i.e. Celo and Arbitrum are going to be fast follows). Can you think through the architecture of the scripts and how we structure the environment variables to make sure we can:

- Quickly deploy to a new chain. Ideally this is as close to just running the script as possible, with the new input data just for that network.
- Quickly redeploy and upgrade to all chains. Ideally this is also as close to just running the script as possible and redeploying the contracts to all supported networks at once.
- Have the same address on every chain for all the proxies. I really want to be able to have a small table in the docs that has the contracts and the addresses and then separately a list of all the chains we currently support.

thelostone-mc (approved):

PR #119, opened 2023-07-23
[WIP] Adding Safe Strategy

codenamejason:


thelostone-mc:
Could we throw in the spec as well ?
 So that we know what to build against

thelostone-mc:
Closing this out for now ! Will reopen when we priotise this

PR #125, opened 2023-07-26
Minimal Anchor Address Implementation

zobront:
Here's a simple implementation of the Anchor.

A few important things to note:
- This should be deployed once and then clones should be put in front of it for each individual deployment.
- Clones should be created with CREATE2 to ensure they can be consistent across chains.
- Clones should use `keccak256(name)` as the salt, so that if an identity changes their name, they get a new anchor address.
- The logic in Registry.sol needs to change so that users remain owners of identities, even after they change the name.

### DIFF
--- a/contracts/core/Registry.sol
+++ b/contracts/core/Registry.sol
@@ -257,15 +257,12 @@ contract Registry is IRegistry, Native, AccessControl, Transfer {
     /// @notice Generates and deploy the anchor for the given identityId and name
     /// @param _identityId Id of the identity
     /// @param _name The name of the identity
-    function _generateAnchor(bytes32 _identityId, string memory _name) internal returns (address) {
+    function _generateAnchor(bytes32 _identityId, string memory _name) internal returns (address anchor) {
         bytes32 salt = keccak256(abi.encodePacked(_identityId, _name));
 
-        bytes memory creationCode = abi.encodePacked(type(Anchor).creationCode, abi.encode(address(this)));
+        bytes memory creationCode = abi.encodePacked(type(Anchor).creationCode, abi.encode(_identityId));
 
-        address anchor = CREATE3.deploy(salt, creationCode, 0);
-        Anchor(anchor).initialize(_identityId);
-
-        return anchor;
+        anchor = CREATE3.deploy(salt, creationCode, 0);
     }
 
     /// @notice Generates the identityId based on msg.sender
--- a/contracts/core/Registry.sol
+++ b/contracts/core/Registry.sol
@@ -140,7 +140,10 @@ contract Registry is IRegistry, Native, AccessControl, Transfer {
         Identity storage identity = identitiesById[_identityId];
         identity.name = _name;
 
-        // set new anchor while preserving old anchor
+        // remove old anchor
+        anchorToIdentityId[identity.anchor] = bytes32(0);
+
+        // set new anchor
         identity.anchor = anchor;
         anchorToIdentityId[anchor] = _identityId;
 
--- a/contracts/core/Registry.sol
+++ b/contracts/core/Registry.sol
@@ -3,10 +3,12 @@ pragma solidity 0.8.19;
 
 // External Libraries
 import {AccessControl} from "@openzeppelin/access/AccessControl.sol";
+import {CREATE3} from "@solady/utils/CREATE3.sol";
 import {ERC20} from "@solady/tokens/ERC20.sol";
 // Interfaces
 import "./IRegistry.sol";
 // Internal Libraries
+import {Anchor} from "./Anchor.sol";
 import {Metadata} from "./libraries/Metadata.sol";
 import "./libraries/Native.sol";
 import "./libraries/Transfer.sol";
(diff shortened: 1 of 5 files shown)


contracts/core/Registry.sol
```diff
@@ -252,12 +251,15 @@ contract Registry is IRegistry, Native, AccessControl, Transfer {
...
     /// @param _name The name of the identity
-    function _generateAnchor(bytes32 _identityId, string memory _name) internal pure returns (address) {
-        bytes32 attestationHash = keccak256(abi.encodePacked(_identityId, _name));
-        return address(uint160(uint256(attestationHash)));
+    function _generateAnchor(bytes32 _identityId, string memory _name) internal returns (address) {
+        bytes32 salt = keccak256(abi.encodePacked(_identityId, _name));
+
+        bytes memory creationCode = abi.encodePacked(type(Anchor).creationCode, abi.encode(_identityId));
+
+        return CREATE3.deploy(salt, creationCode, 0);
```
thelostone-mc, line 262: need to clarify what the salt should be

KurtMerbeth: > need to clarify what the salt should be

i think identityId and name are fine. it's like we had before

thelostone-mc (approved):

contracts/core/Registry.sol
```diff
@@ -138,10 +140,7 @@ contract Registry is IRegistry, Native, AccessControl, Transfer {
         Identity storage identity = identitiesById[_identityId];
         identity.name = _name;
 
-        // clear old anchor
-        anchorToIdentityId[identity.anchor] = bytes32(0);
-
-        // set new anchor
+        // set new anchor while preserving old anchor
         identity.anchor = anchor;
         anchorToIdentityId[anchor] = _identityId;
```
zobront, line 145: Hmm... this is tricky. Let's think through it...
- anchorToIdentityId is saying that a given anchor corresponds with an identity
- once an identity's info changes (ie name change), then the anchor changes too
- this is because we might have attestations saying, for example, "we've verified this anchor has a valid name, because they've shown us their twitter"
- so if the old anchor can point to the new identity, it might keep that attestation, which is a problem

So we can't keep the anchorToIdentityId[identity.anchor] actively pointing to the new identity.

Fortunately, I don't think we have to. If we were looking up the owner of an anchor as (psuedocode) `isOwnerofIdenity(anchorToIdentityId[msg.sender])` then we'd need to solve this. But since the anchor is saving the `identityId` locally, I believe the old code would solve itself.

- Anchor address saves `identityId`
- If a name change happens, the `anchorToIdentityId` is deleted
- But if someone calls execute on it, as long as they're the owner of the `identityId` (which hasn't changed), then it'll still pass
- So everything should be fine

Do you agree?

contracts/core/Registry.sol
```diff
@@ -252,12 +251,18 @@ contract Registry is IRegistry, Native, AccessControl, Transfer {
...
 
-    /// @notice Generates the anchor for the given identityId and name
+    /// @notice Generates and deploy the anchor for the given identityId and name
     /// @param _identityId Id of the identity
     /// @param _name The name of the identity
-    function _generateAnchor(bytes32 _identityId, string memory _name) internal pure returns (address) {
-        bytes32 attestationHash = keccak256(abi.encodePacked(_identityId, _name));
-        return address(uint160(uint256(attestationHash)));
+    function _generateAnchor(bytes32 _identityId, string memory _name) internal returns (address) {
+        bytes32 salt = keccak256(abi.encodePacked(_identityId, _name));
```
zobront, line 258: This is awesome because it's replicable across chains. Same nonce and original msg.sender for creating the identity, same name, and they'll automatically get the same salt. We just need to make sure Registry.sol is at the same address on all chains (which shouldn't be a problem using CREATE2).

zobront (commented):
Looks good! Left a few comments.

zobront:
> > Here's a simple implementation of the Anchor.
> > A few important things to note:
> > 
> > * This should be deployed once and then clones should be put in front of it for each individual deployment.
> > * Clones should be created with CREATE2 to ensure they can be consistent across chains.
> > * Clones should use `keccak256(name)` as the salt, so that if an identity changes their name, they get a new anchor address.
> > * The logic in Registry.sol needs to change so that users remain owners of identities, even after they change the name.
> 
> We talked abut a custom implementation of this https://github.com/ZeframLou/create3-factory
> 
> ~ or ~
> 
> https://github.com/Vectorized/solady/blob/main/src/utils/CREATE3.sol

What's the downside of the way you have it now, with Solady CREATE3?

contracts/core/Registry.sol
thelostone-mc: Ah you are right here ! Preserving the old mapping will cause issues.

thelostone-mc: fixed via 1006b38

contracts/core/Registry.sol
```diff
@@ -252,12 +251,18 @@ contract Registry is IRegistry, Native, AccessControl, Transfer {
...
+
+        bytes memory creationCode = abi.encodePacked(type(Anchor).creationCode, abi.encode(address(this)));
+
+        address anchor = CREATE3.deploy(salt, creationCode, 0);
+        Anchor(anchor).initialize(_identityId);
```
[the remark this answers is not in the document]
thelostone-mc: Makes sense ! Updated this on bf7cef0

thelostone-mc:
no downside on using solady CREATE3. I think. @codenamejason was offering the CREATE3 as alternatives to Clones

thelostone-mc (approved):

PR #140, opened 2023-07-30
Wrapped Voting Spec

0xZakk:
This PR creates a first draft of the spec for Wrapped Voting, a concept we're
hoping to push with protocols to adopt Allo.


KurtMerbeth:
Can we rename the strategy to something more descriptive?
`QFWrappedNftMint` or sth similar? @nfrgosselin @0xZakk

KurtMerbeth:
>The first time they are voted for, they are added as a recipient if the address being voted for is an NFT contract created by the mock protocol factory contract

This is currently not possible in gabewins Mock NFT Protocol POC. Should we stick to his implementation or can we build our own?

thelostone-mc:
This does feel like an example strategy. It is essentially a usecase of having application gating via factory contract. 
And the distribute being a simple transfer to the winner. 
Worth building out to show a usecase BUT this doesn't feel like it should be a production strategy (maybe we could tweak the spec for a more compelling usecase ) @0xZakk ?

0xZakk:
Please feel free to modify the mock NFT "protocol" as much as you need to - that was a quick prototype

0xZakk:
> Can we rename the strategy to something more descriptive?
> WrappedNftMint or sth similar? @nfrgosselin @0xZakk

My preference is to keep the name the same, since it's meant purely as an example of wrapped voting

0xZakk:
> This does feel like an example strategy. It is essentially a usecase of having application gating via factory contract.
> And the distribute being a simple transfer to the winner.
> Worth building out to show a usecase BUT this doesn't feel like it should be a production strategy (maybe we > could tweak the spec for a more compelling usecase ) @0xZakk ?

Yea, the goal here is to have a proof of concept for what this looks like and how it can work. It doesn't have to be "production ready". I opted for a simple example just to show the use case. Most people want to use Allo with QF to do wrapped voting, so if we wanted to beef it up and it wouldn't take too much additional time, then you could implement one of the QF specs for the payout and distribution logic

thelostone-mc:
Yup @KurtMerbeth maybe we throw these in an `prototypeStrategies` folder ?

0xZakk:
Rather than create another folder, can we do this in the header comment in the strategy and in the readme?

For the header comment, I like how OpenZeppelin does these

thelostone-mc:
Closing in favour of #147

PR #141, opened 2023-07-30
Custom (Alt) Registry Spec

0xZakk:
This PR adds a spec for using an alternate registry for recipient eligibility

### DIFF
--- a/contracts/strategies/custom-registry/AltRegistryStrategy.sol
+++ b/contracts/strategies/custom-registry/AltRegistryStrategy.sol
@@ -2,9 +2,8 @@
 pragma solidity 0.8.19;
 
 // External Libraries
-import {SimpleProjectRegistry} from "./mock/SimpleProjectRegistry.sol";
-
-// Intefaces
+import {ReentrancyGuard} from "@openzeppelin/security/ReentrancyGuard.sol";
+// Interfaces
 import {IAllo} from "../../core/IAllo.sol";
 import {IRegistry} from "../../core/IRegistry.sol";
 // Core Contracts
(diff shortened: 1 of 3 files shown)


KurtMerbeth:
Does the recipient need to provide metadata during registration?

> Are there any specific conditions under which the registerRecipient function can or cannot be called?

eg: registration period or sth similar.

> Are multiple registrations for a recipient allowed or just one?
> - What happens in case of a second registration, either in the case of a single registration or multiple registrations?

Is it possible to change the state from Rejected to Approved, or the other way around?

What happens if the pool manager allocates more than 100%?

> Are there any conditions on when the allocate function is callable or not?
eg. after registration period

When is `distribute` callable, or: how to determine if all allocations are set and the contract is ready to payout?

Can we decouple the review process and the allocations, so we can reuse the review logic of donation voting and we don't need to introduce a function like `setIntenalRecipientStatusToInReview`?

In general:
Maybe we can use the `Recipient logic` description of donation voting (just with a custom registry)
 https://github.com/allo-protocol/allo-v2/tree/main/contracts/strategies/donation-voting
 
 
 
Or for the ease of development: can't we just modify/extend one of our existing strategies to use a custom registry?

KurtMerbeth:
@0xZakk 
Since we agreed on using an already existing strategy to implement a custom registry, can this PR be closed?
New Readme: https://github.com/allo-protocol/allo-v2/pull/153/files?short_path=2330242#diff-2330242134251a7ef1bd1af4b76c114ed563edfbc8d9f5f70690bc438619356a

Implementation:
https://github.com/allo-protocol/allo-v2/blob/6331a94ebd159ddb58526f1a14aec9e43b44f326/contracts/strategies/donation-voting-custom-registry/DonationVotingCustomRegistryStrategy.sol

thelostone-mc:
Closing in favour of https://github.com/allo-protocol/allo-v2/pull/153

Issue #137, opened 2023-07-28
privy


nfrgosselin:

KurtMerbeth:
PR: https://github.com/allo-protocol/allo-v2/pull/142

PR #142, opened 2023-07-30
proportional payout strategy

0xZakk:
This PR implements a spec for Proportional Payouts (i.e. the strategy we need
for Givetogether.xyz, the collaboration we're doing with Privy). It also has the
start to an implementation that used our old strategy implementation. Some of
the logic in the contract will be salvageable, but a lot of it will have to
change to fit how we're doing strategies now.


KurtMerbeth:
fixes #137

contracts/core/Allo.sol
```diff
@@ -58,6 +58,7 @@ contract Allo is IAllo, Native, Transfer, Initializable, Ownable, AccessControl
     mapping(address => uint256) private _nonces;
 
     /// @notice Pool.id -> Pool
+    // Note: the ProportionalPayout strategy needs this public?
```
thelostone-mc, line 61: Can't we make it internal ?

KurtMerbeth (approved):

contracts/core/Allo.sol
thelostone-mc:
```suggestion
```

thelostone-mc (approved):

PR #147, opened 2023-08-02
Rapped voting

### LINKED ISSUE #135: rapped voting 🎤

KurtMerbeth:
fixes #135


contracts/strategies/wrapped-voting/solmate/ERC721.sol
```diff
@@ -0,0 +1,207 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity >=0.8.0;
```
thelostone-mc, line 2: are we changing anything on the solady contract ? 
if not, let's use it from the library directly  
if yes -> let's document the changes (like how we did for EAS resolver)

contracts/strategies/IStrategy.sol
```diff
@@ -74,7 +74,7 @@ interface IStrategy {
     function getRecipientStatus(address _recipientId) external view returns (RecipientStatus);
 
     /// @return Input the values you would send to distribute(), get the amounts each recipient in the array would receive
-    function getPayouts(address[] memory _recipientIds, bytes memory _data, address _sender)
+    function getPayouts(address[] memory _recipientIds, bytes[] memory _data)
```
thelostone-mc, line 77: lgtm

contracts/strategies/direct-grants-simple/DirectGrantsSimpleStrategy.sol
```diff
@@ -352,7 +333,7 @@ contract DirectGrantsSimpleStrategy is BaseStrategy, ReentrancyGuard {
             IAllo.Pool memory pool = allo.getPool(poolId);
             allocatedGrantAmount += grantAmount;
 
-            if (allocatedGrantAmount > pool.amount) {
+            if (allocatedGrantAmount > poolAmount) {
```
thelostone-mc, line 336: Okie this makes sense cause poolAmount is updated anytime funding is increased right ? It might be nicer to read the amount always from Allo.sol but that does cost more gas and this instance -> doesn't make sense

contracts/strategies/wrapped-voting/WrappedVotingStrategy.sol
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    /// @notice Internal function to get the payout summary for the recipient
+    /// @param _recipientId The address of the recipient
+    /// @return The PayoutSummary for the recipient
+    function _getPayout(address _recipientId, bytes memory) internal view override returns (PayoutSummary memory) {
+        if (_recipientId != currentWinner) PayoutSummary(_recipientId, 0);
+        return PayoutSummary(currentWinner, poolAmount);
+    }
+}
```
thelostone-mc, line 201: @0xZakk i'd suggest we either have this within an nft folder or update the name to wrappedNFTStrategy. As we add more wrapped strategies, this will lead to confusion and we'd have to rename stuff then which we'd rather avoid 

sidenote @KurtMerbeth thoughts on remove Voting from the name cause the strategy does both voting "allocation" and distribution

thelostone-mc (commented):
Dropped in some feedback. Overall love the direction

contracts/strategies/wrapped-voting/WrappedVotingStrategy.sol
0xZakk: We're only going to have the one POC for wrapped voting and given that it's a POC I think it would be best to keep it called `wrapped-voting` or something like that

codenamejason: `NFTWrappedStrategy.sol` ?

PR #224, opened 2023-08-18
Natspec

### LINKED ISSUE #186: [Major] Add natspec comments
**Timebox to one day**

codenamejason:
This adds NATSPEC comments to our contracts

fixes #186


contracts/core/Allo.sol
```diff
@@ -13,30 +13,29 @@ import {Clone} from "./libraries/Clone.sol";
...
+///     /:/\:\ \:\__\  /:/__/     /:/__/     /:/__/ \:\__\
+///     \/__\:\/:/  /  \:\  \     \:\  \     \:\  \ /:/  /
+///          \::/  /    \:\  \     \:\  \     \:\  /:/  /
+///          /:/  /      \:\  \     \:\  \     \:\/:/  /
+///         /:/  /        \:\__\     \:\__\     \::/  /
+///         \/__/          \/__/      \/__/      \/__/
+///
+/// @notice The Allo core contract
+/// @dev This contract is used to create & manage pools as well as manage the protocol. It
+///      is the core of all things Allo.
```
KurtMerbeth, line 30: `It is the core of all things Allo.`

Does this sentence makes sense?

contracts/core/Allo.sol
```diff
@@ -77,44 +76,66 @@ contract Allo is IAllo, Native, Transfer, Initializable, Ownable, AccessControl
...
-    /// ==== External/Public Functions =====
-    /// ====================================
+    //  ====================================
+    //  ==== External/Public Functions =====
+    //  ====================================
 
     /// @notice Creates a new pool (with custom strategy)
-    /// @param _profileId The profileId of the pool
-    /// @param _strategy The address of strategy
+    /// @dev Must be a member or owner of a profile to create a pool with or without a custom strategy
```
KurtMerbeth, line 124: Who must be a member or owner?

contracts/core/Allo.sol
codenamejason: It needs updated.

contracts/core/Allo.sol
codenamejason: Yeah the msg.sender I'll add that.

contracts/core/Allo.sol
```diff
@@ -77,44 +76,66 @@ contract Allo is IAllo, Native, Transfer, Initializable, Ownable, AccessControl
...
+    ///
+    /// Requirements: The strategy address passed must not be a cloneable strategy
+    ///               The strategy address passed must not be the zero address
+    ///               The caller must be a member or owner of the profile
```
KurtMerbeth, line 128: Suggestion:
The caller must be a member or owner of the `related registry` profile

contracts/core/Allo.sol
```diff
@@ -77,44 +76,66 @@ contract Allo is IAllo, Native, Transfer, Initializable, Ownable, AccessControl
...
+    ///
+    /// @param _profileId The profileId of the pool, used to check if the caller is a member or owner of the profile
```
KurtMerbeth, line 130: Maybe sth like: The profileId of the registry profil.

contracts/core/Allo.sol
```diff
@@ -77,44 +76,66 @@ contract Allo is IAllo, Native, Transfer, Initializable, Ownable, AccessControl
...
     /// @param _initStrategyData The data to initialize the strategy
-    /// @param _token The address of the token
-    /// @param _amount The amount of the token
-    /// @param _metadata The metadata of the pool
-    /// @param _managers The managers of the pool
+    /// @param _token The address of the token you want to use in your pool
+    /// @param _amount The amount of the token you want to deposit into the pool on initialization
+    /// @param _metadata The metadata of the pool, this uses our 'Meatdata.sol' struct (consistent throughout the protocol)
+    /// @param _managers The managers of the pool, and can be added/removed later by the pool admin
+    /// @custom:initstrategydata The encoded data will be specific to a given strategy requirements, reference the strategy implementation of 'initialize()'
```
KurtMerbeth, line 137: `@custom:_initStrategyData ....`

contracts/core/Allo.sol
KurtMerbeth: the comment from `createPool` is nice:
`/// Requirements: 'msg.sender' must be owner or member of the profile id passed as '_profileId'`

contracts/core/Allo.sol
```diff
@@ -177,36 +210,51 @@ contract Allo is IAllo, Native, Transfer, Initializable, Ownable, AccessControl
     }
 
     /// @notice Updates the registry address
+    /// @dev Use this to update the registry address
+    ///
+    /// Requirements: The caller must be allo owner
```
KurtMerbeth, line 215: `Allo` instead of `allo`, since `Allo` is a name

contracts/core/Allo.sol
```diff
@@ -216,101 +264,155 @@ contract Allo is IAllo, Native, Transfer, Initializable, Ownable, AccessControl
...
+    ///
     /// @param _poolId The pool id
     /// @param _manager The address remove
     function removePoolManager(uint256 _poolId, address _manager) external onlyPoolAdmin(_poolId) {
         _revokeRole(pools[_poolId].managerRole, _manager);
     }
 
     /// @notice Transfer thefunds recovered  to the recipient
+    ///
+    /// Requirements: The caller must be a pool owner
```
KurtMerbeth, line 310: only Allo owner, not pool owner

contracts/core/Allo.sol
KurtMerbeth: Shouldn't we maybe use `@dev` or `@notice` instead of `@custom`?

contracts/core/Allo.sol
```diff
@@ -216,101 +264,155 @@ contract Allo is IAllo, Native, Transfer, Initializable, Ownable, AccessControl
...
-    /// @param _data encoded data unique to strategy for each pool
+    /// @notice Register multiple recipients to multiple pools
+    /// @dev Returns the recipientIds from the strategy that have been registered from calling this funciton
+    ///
+    /// Requirements: Encoded '_data' length must match _poolIds length or this will revert with MISMATCH()
+    ///               Other requirements will be determined by the strategy
+    ///
+    /// @param _poolIds Id of the pools
+    /// @param _data Encoded data unique to a strategy that registerRecipient() requires
+    /// @custom:data The encoded data will be specific to a given strategy requirements, reference the strategy implementation of registerRecipient()
```
KurtMerbeth, line 347: Shouldn't we maybe use `@dev` or `@notice` instead of `@custom`?

contracts/core/Allo.sol
```diff
@@ -216,101 +264,155 @@ contract Allo is IAllo, Native, Transfer, Initializable, Ownable, AccessControl
...
     {
         uint256 poolIdLength = _poolIds.length;
-        address[] memory recipientIds = new address[](poolIdLength);
+        recipientIds = new address[](poolIdLength);
 
         if (poolIdLength != _data.length) {
             revert MISMATCH();
         }
 
+        // Loop through the _poolIds & _data and call the internal _registerRecipient() function
```
KurtMerbeth, line 360: `strategy.registerRecipient()` and not the internal `_registerRecipient()`

contracts/core/Allo.sol
```diff
@@ -216,101 +264,155 @@ contract Allo is IAllo, Native, Transfer, Initializable, Ownable, AccessControl
...
         _allocate(_poolId, _data);
     }
 
     /// @notice vote to multiple pools
+    ///
+    /// Requirements: This will be determined by the strategy
+    ///
     /// @param _poolIds ids of the pools
     /// @param _datas encoded data unique to the strategy for that pool
+    /// @custom:datas The encoded data will be specific to a given strategy requirements, reference the strategy implementation of allocate()
```
KurtMerbeth, line 406: Shouldn't we maybe use `@dev` or `@notice` instead of `@custom`?

contracts/core/Allo.sol
```diff
@@ -320,8 +422,12 @@ contract Allo is IAllo, Native, Transfer, Initializable, Ownable, AccessControl
     }
 
     /// @notice passes _data & msg.sender through to the disribution strategy for that pool
+    ///
+    /// Requirements: This will be determined by the strategy
+    ///
     /// @param _poolId id of the pool
     /// @param _data encoded data unique to the strategy for that pool
+    /// @custom:data The encoded data will be specific to a given strategy requirements, reference the strategy implementation of distribute()
```
KurtMerbeth, line 430: Shouldn't we maybe use `@dev` or `@notice` instead of `@custom`?

contracts/core/Allo.sol
```diff
@@ -331,13 +437,15 @@ contract Allo is IAllo, Native, Transfer, Initializable, Ownable, AccessControl
     /// ====================================
 
     /// @notice Creates a new pool
+    /// @dev Must be a member or owner of a profile to create a pool
```
KurtMerbeth, line 440: who must be a member or owner?

contracts/core/Allo.sol
```diff
@@ -419,7 +530,7 @@ contract Allo is IAllo, Native, Transfer, Initializable, Ownable, AccessControl
         pools[_poolId].strategy.allocate{value: msg.value}(_data, msg.sender);
     }
 
-    /// @notice Deduct the fee and transfers the amount to the distribution strategy
+    /// @notice Deducts the fee and transfers the amount to the distribution strategy
```
KurtMerbeth, line 533: `@dev Deducts the fee and transfers the amount to the distribution strategy`
`@notice Fund the pool ..`

contracts/core/Allo.sol
```diff
@@ -530,32 +648,32 @@ contract Allo is IAllo, Native, Transfer, Initializable, Ownable, AccessControl
         return address(pools[_poolId].strategy);
     }
 
-    /// @notice return fee percentage
+    /// @notice Getter for fee percentage
```
KurtMerbeth, line 651: missing @return

contracts/core/Allo.sol
```diff
@@ -530,32 +648,32 @@ contract Allo is IAllo, Native, Transfer, Initializable, Ownable, AccessControl
...
     function getPercentFee() external view returns (uint256) {
         return percentFee;
     }
 
-    /// @notice return base fee
+    /// @notice Getter for base fee
```
KurtMerbeth, line 656: missing @return

contracts/core/Allo.sol
```diff
@@ -530,32 +648,32 @@ contract Allo is IAllo, Native, Transfer, Initializable, Ownable, AccessControl
...
     function getBaseFee() external view returns (uint256) {
         return baseFee;
     }
 
-    /// @notice return treasury
+    /// @notice Getter for treasury address
```
KurtMerbeth, line 661: missing @return

contracts/core/Allo.sol
```diff
@@ -530,32 +648,32 @@ contract Allo is IAllo, Native, Transfer, Initializable, Ownable, AccessControl
...
     function getTreasury() external view returns (address payable) {
         return treasury;
     }
 
-    /// @notice return registry
+    /// @notice Getter for registry
```
KurtMerbeth, line 666: missing @return

contracts/core/Allo.sol
```diff
@@ -530,32 +648,32 @@ contract Allo is IAllo, Native, Transfer, Initializable, Ownable, AccessControl
...
     function getRegistry() external view returns (IRegistry) {
         return registry;
     }
 
-    /// @notice return boolean if strategy is cloneable
+    /// @notice Getter for if strategy is cloneable
```
KurtMerbeth, line 671: missing @return

contracts/core/Allo.sol
```diff
@@ -530,32 +648,32 @@ contract Allo is IAllo, Native, Transfer, Initializable, Ownable, AccessControl
...
     function isCloneableStrategy(address _strategy) external view returns (bool) {
         return _isCloneableStrategy(_strategy);
     }
 
-    /// @notice return the pool
+    /// @notice Getter for the 'Pool'
```
KurtMerbeth, line 676: missing @param and @return

contracts/core/IAllo.sol
```diff
@@ -45,55 +64,180 @@ interface IAllo {
         uint256 amount,
         Metadata metadata
     );
+
+    // @dev Event emitted when a pools metadata is updated
```
KurtMerbeth, line 68: should we add params to the events? If yes:  need to be changed globally

contracts/core/IAllo.sol
```diff
@@ -45,55 +64,180 @@ interface IAllo {
...
+    // @dev Initialize the Allo contract
+    //
+    // Requirements:
+    //
     function initialize(address _registry, address payable _treasury, uint256 _percentFee, uint256 _baseFee) external;
 
+    // @dev Updates the pools metadata
+    //
+    // Requirements: 'msg.sender' must be a pool admin
+    //
```
KurtMerbeth, line 108: should we add `@param` to interface functions?
we could c&p from the implementation

contracts/core/Allo.sol
KurtMerbeth: should we also mention that this function is NOT payable?

contracts/core/IRegistry.sol
```diff
@@ -21,49 +31,149 @@ interface IRegistry {
...
 
     /// ======================
     /// ======= Events =======
     /// ======================
 
+    /// @dev Event emitted when a profile is created
+    ///
+    /// Note: This will return your anchor address
+    ///
+
```
KurtMerbeth, line 59: whole interface misses `@param` and `@return`

contracts/core/Registry.sol
```diff
@@ -14,8 +14,8 @@ import "./libraries/Native.sol";
 import "./libraries/Transfer.sol";
 
 /// @title Registry
-/// @notice Registry contract for profiles
-/// @dev This contract is used to create and manage profiles
+/// @notice Registry contract for creating and managing profiles
```
KurtMerbeth, line 17: should we already mention the Anchor here?

contracts/core/Registry.sol
```diff
@@ -46,10 +46,16 @@ contract Registry is IRegistry, Native, AccessControl, Transfer {
         _;
     }
 
+    /// @notice Contract constructor
+    ///
```
KurtMerbeth, line 50: `@dev Reverts if _owner is address(0)`

contracts/core/Registry.sol
```diff
@@ -106,125 +127,176 @@ contract Registry is IRegistry, Native, AccessControl, Transfer {
...
+        // Emit the event that the profile was created
         emit ProfileCreated(profileId, profile.nonce, profile.name, profile.metadata, profile.owner, profile.anchor);
 
+        // Return the profileId
         return profileId;
     }
 
     /// @notice Updates the name of the profile and generates new anchor
+    ///
+    /// Requirements: Must be the owner of the profile
```
KurtMerbeth, line 156: who must be owner of the profile?

contracts/core/Registry.sol
```diff
@@ -106,125 +127,176 @@ contract Registry is IRegistry, Native, AccessControl, Transfer {
...
 
+        // Get the profile using the profileId from the mapping
         Profile storage profile = profilesById[_profileId];
+
+        // Set the new name
         profile.name = _name;
 
-        // remove old anchor
+        // Remove old anchor
+        // TODO: Why do we set to address zero to only update again with new anchor? @thelostone-mc @kurtmerbeth
```
KurtMerbeth, line 180: we do not update _with_ the anchor again

Its a anchor => profile mappping

when creating a new anchor, the `old anchor` loses the link to the profile and the the link is set to the `new anchor`

contracts/factory/ContractFactory.sol
```diff
@@ -1,28 +1,91 @@
...
+ * @dev ContractFactory is used internally to deploy our contracts using CREATE3
+ *
+ */
 contract ContractFactory {
+    /// ======================
+    /// ======= Errors =======
+    /// ======================
+
+    /**
+     * Returns when the requested salt has already been used
```
KurtMerbeth, line 19: Reverts

contracts/factory/ContractFactory.sol
```diff
@@ -1,28 +1,91 @@
...
+     */
     error SALT_USED();
+
+    /**
+     * Returns when the caller is not authorized to deploy
```
KurtMerbeth, line 24: Reverts

contracts/factory/ContractFactory.sol
```diff
@@ -44,7 +107,15 @@ contract ContractFactory {
         emit Deployed(deployedContract, salt);
     }
 
+    /**
+     * @notice Set the allowed deployer
+     * @dev Sets the '_deployer' to '_allowedToDeploy'
```
KurtMerbeth, line 112: to add: only deployer

contracts/strategies/direct-grants-simple/DirectGrantsSimpleStrategy.sol
```diff
@@ -73,44 +108,74 @@ contract DirectGrantsSimpleStrategy is BaseStrategy, ReentrancyGuard {
...
+
+    /// @notice This maps accepted recipients to their upcoming milestone
+    /// @dev Mapping of the 'recipientId' to the 'nextMilestone' index
     mapping(address => uint256) public upcomingMilestone;
-    /// @notice recipientId -> Total Milestones for the recipient
-    mapping(address => uint256) public totalMilestones;
+
+    /// @notice This maps accepted recipients to their total milestones
+    // TODO: (payouts or count?) @thelostone-mc @KurtMerbeth
+    // NOTE: Were not using this in the contract, but it is here for future use?
```
KurtMerbeth, line 128: can be removed

contracts/strategies/direct-grants-simple/DirectGrantsSimpleStrategy.sol
```diff
@@ -73,44 +108,74 @@ contract DirectGrantsSimpleStrategy is BaseStrategy, ReentrancyGuard {
...
     constructor(address _allo, string memory _name) BaseStrategy(_allo, _name) {}
 
     /// ===============================
     /// ========= Initialize ==========
     /// ===============================
 
+    /// @notice Initialize the strategy
+    /// @param _poolId Id of the pool
+    /// @param _data The data to be decoded
+    /// @custom:data (bool registryGating, bool metadataRequired, bool grantAmountRequired)
```
KurtMerbeth, line 149: i like the format

contracts/strategies/donation-voting-merkle-distribution/DonationVotingMerkleDistributionStrategy.sol
```diff
@@ -27,13 +31,29 @@ contract DonationVotingMerkleDistributionStrategy is BaseStrategy, ReentrancyGua
...
+    /// The 'rowIndex' is the index of the row in the bitmap, and the 'statusRow' is the value of the row.
+    /// The 'statusRow' is updated when the status of a recipient changes.
+    ///
+    /// Note: Since we need 4 bits to store a status, one row of the bitmap can hold the status information of 256/4 recipients.
+    ///
+    /// For example, if we have 5 recipients, the bitmap will look like this:
+    /// | recipient1 | recipient2 | recipient3 | recipient4 | recipient5 | 'rowIndex'
+    /// |     0000   |    0001    |    0010    |    0100    |    1000    | 'statusRow'
+    /// |     none   |   pending  |  accepted  |  rejected  |  appealed  | converted status (0, 1, 2, 4, 8)
+    ///
```
KurtMerbeth, line 48: | status | int | bin |
| -- | -- | -- |
| none | 0 | 0000 |
| pending | 1 | 0001 |
| accepted | 2 | 0010 |
| -- | -- | -- |
| **rejected** | **3** | **0011** |
| **appealed** | **4** | **0100** |

contracts/strategies/donation-voting-merkle-distribution/DonationVotingMerkleDistributionStrategy.sol
```diff
@@ -56,100 +77,183 @@ contract DonationVotingMerkleDistributionStrategy is BaseStrategy, ReentrancyGua
...
+    /// Instead of using 1 bit for each recipient status, we will use 4 bits for each status
+    /// to allow 5 statuses:
+    /// 0: none
+    /// 1: pending
+    /// 2: accepted
+    /// 3: rejected
+    /// 4: appealed
+    ///
+    /// Since it's a mapping the storage it's pre-allocated with zero values, so if we check the
+    /// status of an existing recipient, the value is by default 0 (pending)
```
KurtMerbeth, line 210: default 0 is none, not pending

contracts/strategies/donation-voting-merkle-distribution/DonationVotingMerkleDistributionStrategy.sol
```diff
@@ -328,25 +487,33 @@ contract DonationVotingMerkleDistributionStrategy is BaseStrategy, ReentrancyGua
         _transferAmount(pool.token, msg.sender, _amount);
     }
 
-    /// @notice Claim allocated tokens
+    // TODO: This naming seems a bit off to me, but I'm not sure what a better name would be? @thelostone-mc @KurtMerbeth WDYT?
```
KurtMerbeth, line 490: i don't have any better idea tbh. I'm fine with claim, but not married to it

0xZakk:
Hey y'all, I had a chance to look through the nat spec and it all looks pretty good! Thank you Jason and Kurt for the push here.

I wanted to call out a couple of stylistic things I notice we're doing that will clean up our natspec a lot. This is super important because it will make the lives of the auditors a lot easier if these are done well, which means we (a) won't get complaints from them about it and (b) will get better, more meaningful bug/vulnerability reports from them. While most of these points are functional, some are stylistic. 

### 1. Authors

What I'd like to propose is that for the core contracts, we list all of our names and emails as the authors (just a comma separated list) and then for the strategies we list the names of everyone who contributed to those strategies. I suggest we do names and emails (i.e. `@author 0xZakk <zakk@gitcoin.co>, Nate <nate@gitcoin.co>`) as this is what I've seen the most. I have no preference when it comes to name order, but let's use the same thing for every core contract. 

### 2. Contract-level Nat Spec

Right before the first line of the contract (`contract Whatever {`) we should have: `@title`, `@author`, `@notice`, and `@dev` (in that order). The title should just be the name of the contract. I'll have a point below about the difference between `@notice` and `@dev`.

### 3. Notice and Dev

- `@notice` should explain the business use case of the contract, variable, function, etc. If you were going to explain this to @nfrgosselin, what would you say?
- `@dev` should explain the technical use of and/or any considerations for using this contract, variable, function, etc. if you were going to explain this to me or Zach O, what would you say?
- Every contract, variable, function, error, event, etc - everything needs a `@notice`. Not everything needs an `@dev` - only if there's something not covered in the `@notice` description or a gotcha/nuance that devs specifically need to be aware of.
- Both the `@notice` and the `@dev` should be full-sentences. They can and should be short though!

### 4. Content for notice, dev, param, etc

Please don't use the name of the thing in the definition of the thing. Please also don't put the type of the thing in the definition of the thing.

A common pattern I see in our contracts is something like this:

```solidity
/// @notice Base fee
uint256 internal baseFee;
```

The variable is called `baseFee`, so having a notice that says `Base fee` doesn't add anything that isn't already in the code. (See above on the purpose of `@notice` and `@dev`)

The same goes for types: if something is of type `address`, then you don't need to include that it's an address in the description. That context is already part of the code.

This also applies with our use of `@param` too, for example:

```solidity
/// @param _profileId The profileId of the allowed profile to execute calls
```

The param is called `_profileId`, so the description needs to explain what that is and/or why it's there, ideally without reference to the name (`profileId`). Here, you could say something like: `The ID of the profile allowed to execute calls`

And this goes for everything, including errors and events:

```solidity
/// @dev Event emitted when a new pool is created
```

The next line is `event PoolCreated( ... )`, so the description doesn't need to communicate that this is an event.

### 5. Events and Errors

- The `@notice` needs to say when this is emitted or thrown (the sentence should ideally use the word "emitted" or "thrown")
- We should use a template sentence for every error and event across all the contracts so they're consistent from contract to contract (i.e. `@notice Emitted when a pool is created` or `@notice Thrown when the user is not authorized to perform a certain action`

### 6. `//`, `/* */`, and `///`

We need to use `///` for every nat spec line and then either `//` or `/* */` for everything else. Some of the nat spec comments are using `//` and some non-natspec things (like the Allo asci-art) are using `///`

I know you can use `/* */` with nat spec, but it's a lot cleaner to just keep everything consistent to `///`

All of this gets used to generate contract-docs, so we want the formatting to be good to avoid any weird rendering when we generate the contract docs

### Misc other notes:

- Spacing: There shouldn't be any empty lines between the natspec comments and the line of code that it's commenting on (i.e. in `IRegistry.sol`). Nor should there be any empty lines in between different nat spec comments (i.e. in `Registry.sol`)
- Probably everything we have listed as `Requirements: ` should really just be part of the `@dev` comment
- Write comments using markdown. That means that if something is a piece of code, surround it with backticks instead of single or double quotes.

Okay, I know this is a lot but hopefully this is helpful! I think we're like 80% of the way there on these and I really appreciate your attention to detail on these.

thelostone-mc (approved):

PR #230, opened 2023-08-21
chore: update milestone logic

### LINKED ISSUE #214: [Major] DirectGrantSimpleStrategy: Ensure we cannot assign more than 100% to the milestones
Comment by Zach:
In DirectGrantSimpleStrategy.sol, I think we don’t update upcomingMilestone[recipient] until a distribution is made. But that allows us to call setMilestones() twice, which means we could allocate over 100% and allow some people to double withdraw while others are left with none.

Possible solution: always interate through ALL milestones and check if we exceed 100%

### LINKED ISSUE #227: Direct grants milestone updates
The current direct grants strategy requires the pool manager to set the payout milestones, but the GS team has indicated they want the flow to change so that recipients propose milestones and pool managers approve.

Changes to make:
- Pool manager OR recipient can call `setMilestones` and set the payout gates
    - If recipient calls `setMilestones`, then the pool manager must approve the milestones
    - If pool manager calls `setMilestones`, then the milestones are automatically set

NOTE — if the option logic creates too much scope, we can limit this to just being the recipient who can call `setMilestones`

thelostone-mc:
### Description

The milestone flow is now updated to 

- recipient registers (no changes here)
- pool manager reviews the recipient (no changes here)
- recipient sets milestones (by invoking setMilestones, which can be invoked multiple times until approved)
- pool manager approves milestones (this will be a new function)
- pool manager can also invoke setMilestones to set milestones and approve them

fixes #214 
fixes #227

### DIFF
--- a/contracts/strategies/direct-grants-simple/DirectGrantsSimpleStrategy.sol
+++ b/contracts/strategies/direct-grants-simple/DirectGrantsSimpleStrategy.sol
@@ -31,6 +31,8 @@ contract DirectGrantsSimpleStrategy is BaseStrategy, ReentrancyGuard {
         uint256 grantAmount;
         Metadata metadata;
         InternalRecipientStatus recipientStatus;
+        RecipientStatus milestonesReviewStatus;
+        address creator;
     }
 
     /// @notice Struct to hold milestone details
@@ -62,6 +64,7 @@ contract DirectGrantsSimpleStrategy is BaseStrategy, ReentrancyGuard {
     event MilestoneSubmitted(address recipientId, uint256 milestoneId, Metadata metadata);
     event MilestoneStatusChanged(address recipientId, uint256 milestoneId, RecipientStatus status);
     event MilestonesSet(address recipientId);
+    event MilestonesReviewed(address recipientId, RecipientStatus status);
 
     /// ================================
     /// ========== Storage =============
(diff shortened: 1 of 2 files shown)


contracts/strategies/direct-grants-simple/DirectGrantsSimpleStrategy.sol
```diff
@@ -166,38 +169,42 @@ contract DirectGrantsSimpleStrategy is BaseStrategy, ReentrancyGuard {
...
-            totalAmountPercentage += milestone.amountPercentage;
-            milestones[_recipientId].push(milestone);
+        if (recipient.milestonesReviewStatus == RecipientStatus.Accepted) {
+            revert MILESTONES_ALREADY_SET();
+        }
 
-            unchecked {
-                i++;
-            }
+        bool isRecipientCreator = recipient.creator == msg.sender;
```
KurtMerbeth, line 183: Since the recipientId can only be the sender OR a pool manager, the `creator` property is not needed.
Instead of, we could implement `isRecipientCreator` like that:
`bool isRecipientCreator = ( msg.sender == _recipientId )|| _isProfileMember(recipientId, msg.sender);`

contracts/strategies/direct-grants-simple/DirectGrantsSimpleStrategy.sol
```diff
@@ -319,7 +326,9 @@ contract DirectGrantsSimpleStrategy is BaseStrategy, ReentrancyGuard {
             useRegistryAnchor: registryGating ? true : isUsingRegistryAnchor,
             grantAmount: grantAmount,
             metadata: metadata,
-            recipientStatus: InternalRecipientStatus.Pending
+            recipientStatus: InternalRecipientStatus.Pending,
+            milestonesReviewStatus: RecipientStatus.Pending,
+            creator: _sender
```
KurtMerbeth, line 331: can be reoved. see other comment

KurtMerbeth:
@thelostone-mc 

Check out https://github.com/allo-protocol/allo-v2/pull/230/commits/91b5639d54aac9c410d4c7e13ebaf466f2b1b89a
I added a comment and a test case, which fails currently. We need to improve the code so the test case does not fail anymore

KurtMerbeth (approved):

PR #237, opened 2023-08-22
Update RFPSimpleStrategy.sol: Enhanced Gas Optimizations and fixed some errors in comments

Aryan9592:
Here's the explanation of why I changed the order of those particular state variables:

Usually, EVM stores state variables or storage variables in slots, each slot upto 32 bytes of data and these slots are directly related to gas, i.e more the slots, more the gas gonna be used up!!

So if by anyway we reduced the number of slots, we might optimize the gas a little bit. Take a look at this:

Screenshot (15)

Just for some information sake:

- Address takes up 20 bytes
- bool takes up 1 byte
- uint256 takes up 32 bytes and covers up the whole storage

and here's my implementation:

Screenshot (18)

Saving 1 extra slot, Hope that makes you clear!!

Plus there's some grammatical fixes in the comments, just some extra stuff done...

### DIFF
--- a/contracts/strategies/rfp-simple/RFPSimpleStrategy.sol
+++ b/contracts/strategies/rfp-simple/RFPSimpleStrategy.sol
@@ -16,7 +16,7 @@ contract RFPSimpleStrategy is BaseStrategy, ReentrancyGuard {
     /// ========== Storage =============
     /// ================================
 
-    /// @notice Struct to hold details of an recipient
+    /// @notice Struct to hold details of a recipient
     struct Recipient {
         bool useRegistryAnchor;
         address recipientAddress;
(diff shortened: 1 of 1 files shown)


contracts/strategies/rfp-simple/RFPSimpleStrategy.sol
```diff
@@ -16,7 +16,7 @@ contract RFPSimpleStrategy is BaseStrategy, ReentrancyGuard {
     /// ========== Storage =============
     /// ================================
 
-    /// @notice Struct to hold details of an recipient
+    /// @notice Struct to hold details of a recipient
```
codenamejason, line 19: Don't worry about the natspec here, we have #224 to take care of this.

Aryan9592: So I should change it back to the original, or just keep it as it is.
Moreover, it kind of looks grammatically good. and if you want me to change it, won't be a big deal !!

thelostone-mc (approved):

0xKurt (approved):

thelostone-mc:
Thanks for the PR @Aryan9592 ^_^

Issue #223, opened 2023-08-17
Fix Allo v2 deploy script


0xKurt:
After deploying the contracts, the Allo contract is not initialized. We need to fix it.

- [x] PGN and Celo verifications need to be checked for all contracts
- [x] PGN and Celo Allo initialization needs to be checked
- [x] PGN and Celo Whitelist of Merkle and Direct strategies needs to be checked
- [x] Deploy script needs to be updated to init the allo contract after deployment
- [x] Add documentation / requirement md file for chain requirements to deploy (EVM, hardhat-upgradeable, gnosis safe, subgraph, permit2, anything else?)

nfrgosselin:
This is finished per standup on 9/6

PR #243, opened 2023-08-23
fix import paths

### LINKED ISSUE #236: OpenZeppelin import issues
I think the issue has to do with submodules and how Forge handles these. I suspect this just isnt playing nicely with the open zeppelin contracts, which are installed with Forge and through NPM.

0xZakk:
Fixes #236

I've added the `eas-contracts` with Foundry. I've also changed the EAS and OpenZeppelin import paths to pull from `lib/` instead of from `node_modules/`. This is what was causing errors when someone installed the `allo-v2` repo using Foundry.

Commits:
- add and configure hardhat preprocessor
- forge install: eas-contracts
- update import paths to pull from foundry's lib/ and not from node_modules

Flagging that it looks like the build scripts are broken or not documented correctly? What's in the `DEPLOY_STEPS.md` doc is different from what's in the npm scripts in `package.json` is also different from what's in the `scripts/` directory. Can we add that to #223?


0xKurt:
@thelostone-mc @0xZakk does the test fail for you too? I can't reproduce the failing test locally and therefore i can't look into fixing it.

0xZakk:
@0xKurt should be good now

0xZakk:
@0xKurt all set

0xKurt (approved):
lgtm!

0xKurt (approved):

0xKurt (approved):

PR #273, opened 2023-08-30
Zk spike

codenamejason:
Success on v2 zk era testnet


thelostone-mc:
i believe the zk-era spike was for v1 ?

0xKurt:
> i believe the zk-era spike was for v1 ?

for both

Issue #259, opened 2023-08-25
[spike] scope out pre-audit for common token issues


nfrgosselin:
**timebox to half a day (at most)**

- pull together list of common issues with tokens (i.e. https://github.com/d-xo/weird-erc20)
- create ticket to address those issues

thelostone-mc:
Questions:

- ✅ Reentrant Calls : we have the the nonReentrant modifier on all token transfer calls
- ✅ Missing Return Values: We use safeTransfer so that should handle bad tokens where return value isn't implemented right
- Fee on Transfer
- Balance Modifications Outside of Transfers (rebasing/airdrops)
-❓ Upgradable Tokens : This restriction can be imposed on the frontend. Implementing this on the contract level would require us to keep an rejectedToken list which   
- ✅ Flash Mintable Tokens: Not an issue as we stored funds 
- ✅  Tokens with Blocklists: The most we can do is stay compliant to laws which our protocol does as it's a public goods protocol 
- ✅ Pausable Tokens: We cannot control this and all the token transfer on our contract would be blocked when the token is paused and can be resumed only when unpaused
- Approval Race Protections
-  Revert on Approval To Zero Address: We implement zero address checks 
- Revert on Zero Value Approvals
- ✅ Revert on Zero Value Transfers : We implement zero value check 
- ✅ Multiple Token Addresses ( this would be on the user using the token)
- Low Decimals
- High Decimals
- transferFrom with src == msg.sender
- Non string metadata
- ✅ Revert on Transfer to the Zero Address We implement zero address check 
- ✅ No Revert on Failure : Handled by safeTransfer 
- Revert on Large Approvals & Transfers
- Code Injection Via Token Name
- Unusual Permit Function

PR #274, opened 2023-08-30
reentrancy check for withdraw

thelostone-mc:
Refs #259


0xKurt:
@thelostone-mc do you think this one is needed?
since the poolManager can just withdraw. I think no rrentrancy check is needed here.

PR #288, opened 2023-08-31
Update Allo.sol: Optimized code

Aryan9592:
The changes I made are:

`uint256 i = 0;` -> `uint256 i;`

Why to waste gas for just initializing any variable to 0 even after we know that default value of all `uint` and `int` is 0..

There's one more thing I would love to point out, see one of the code snippets from this file:
```
for (uint256 i = 0; i < numPools;) {
     _allocate(_poolIds[i], _datas[i]);
     unchecked {
        i++;
     }
}
```
Must say, that `unchecked` part is really worth of appreciation and the idea behind it. But what if we make `i++` -> `++i`.
if that `i` won't be used anywhere after this loop, it be safe to use `++i` and will save even more gas. Even I first thought that what will even happen after making this kind of change, but after running a demo code in remix...it was effective.
I haven't made this change yet, that's why pointed it out here cuz logically change in `i++` to `++i` can bring some undesired working of our contract, just you need to look them out and it's worth changing to `++i` if it doesn't make any faulty change. Just let me know

### DIFF
--- a/contracts/core/Allo.sol
+++ b/contracts/core/Allo.sol
@@ -331,7 +331,7 @@ contract Allo is IAllo, Native, Transfer, Initializable, Ownable, AccessControl,
         }
 
         // Loop through the '_poolIds' & '_data' and call the 'strategy.registerRecipient()' function
-        for (uint256 i = 0; i < poolIdLength;) {
+        for (uint256 i; i < poolIdLength;) {
             recipientIds[i] = pools[_poolIds[i]].strategy.registerRecipient(_data[i], msg.sender);
             unchecked {
                 i++;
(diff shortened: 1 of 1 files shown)


codenamejason (approved):

Aryan9592:
what about `i++` to `++i`? Does this change making any undesired behaviour in logic of the code?

thelostone-mc:
@Aryan9592 you are right, we could.
Since the increment happens at the end , it should be good to do that.
Would you want to submit another PR for that ?

```
for (uint256 i = 0; i <= numPools;) {
     _allocate(_poolIds[i], _datas[i]);
     unchecked {
       ++ i;
     }
}
```

thelostone-mc (approved):

Aryan9592:
@thelostone-mc Submitting another PR won't be a big deal, but this change could be held in many files..Cuz that's not the only loop we gonna fix up right?
Let me submit a new PR fixing this in the current file i.e `Allo.sol`, if you feel that in any particular loop `++i` is causing some problems, we will remove it

PR #298, opened 2023-09-01
PERMIT2

codenamejason:


test/foundry/strategies/DonationVotingMerkleDistributionBase.t.sol
```diff
@@ -993,4 +999,32 @@ contract DonationVotingMerkleDistributionBaseMockTest is
         //   '0x4a3e9be6ab6503dfc6dd903fddcbabf55baef0c6aaca9f2cce2dc6d6350303f5'
         // ]
     }
+
+    function __getPermitTransferSignature(
+        ISignatureTransfer.PermitTransferFrom memory permit,
+        uint256 privateKey,
+        bytes32 domainSeparator,
+        address senderStrategy
```
thelostone-mc, line 1007: Ah this is what we had to update

PR #326, opened 2023-09-22
chore: profileId is generated using owner

thelostone-mc:
This enables 

- creating profiles on behalf of other users
- restrict member creation to only profile owner and not creator of profile 

fixes: https://github.com/allo-protocol/allo-v2/issues/327
refs https://github.com/sherlock-audit/2023-09-Gitcoin-judging/issues/7

This would enable GiveEth to optionally create profiles on behalf of the profile owner but will hold no ownership of the profile

### DIFF
--- a/contracts/core/Registry.sol
+++ b/contracts/core/Registry.sol
@@ -114,7 +114,7 @@ contract Registry is IRegistry, Native, AccessControl, Transfer, Initializable, 
     /// @param _name The name of the profile
     /// @param _metadata The metadata of the profile
     /// @param _owner The owner of the profile
-    /// @param _members The members of the profile
+    /// @param _members The members of the profile (can be set only if msg.sender == _owner)
     /// @return The ID for the created profile
     function createProfile(
         uint256 _nonce,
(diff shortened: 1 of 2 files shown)


contracts/core/Registry.sol
```diff
@@ -124,7 +124,7 @@ contract Registry is IRegistry, Native, AccessControl, Transfer, Initializable,
         address[] memory _members
     ) external returns (bytes32) {
         // Generate a profile ID using a nonce and the msg.sender
-        bytes32 profileId = _generateProfileId(_nonce);
+        bytes32 profileId = _generateProfileId(_nonce, _owner);
```
thelostone-mc, line 127: Worth noting:

- A profile ID is not regenerated when the owner changes
- This means if a profile owner changes, while the new profile owner can recreate the same profile anchor on a new chain, they will not be able to control the profile as only the owner (aka the old owner) would have access to it.

This exists in our current architecture as well as the msg.sender has to be the same across chains to ensure your project has the same profile Id and anchor address
In that sense, this is a more logical step as it give you the ability to create profiles on behalf of someone else but not do anything beyond that

contracts/core/Registry.sol
```diff
@@ -147,6 +147,12 @@ contract Registry is IRegistry, Native, AccessControl, Transfer, Initializable,
 
         // Assign roles for the profile members
         uint256 memberLength = _members.length;
+
+        // Only profile owner can add members
+        if (memberLength > 0 && _owner != msg.sender) {
```
thelostone-mc, line 152: This ensures the msg.sender cannot add members on behalf of the owner and this action is restricted only to the profile owner

0xKurt (approved):

codenamejason (approved):

PR #328, opened 2023-09-25
Deploy scripts for latest contracts

codenamejason:


thelostone-mc (changes requested):
@codenamejason everything looks good! 
We just need to update the script folder to ensure the strategies are present in it's own folder so that the scrips are not confusing

thelostone-mc (dismissed):

thelostone-mc (dismissed):

thelostone-mc (approved):

PR #345, opened 2023-10-03
fix: add withdraw function to QVBaseStrategy

thelostone-mc:
refs https://github.com/sherlock-audit/2023-09-Gitcoin-judging/issues/446

In other strategies, poolAmount is reduced as the pot decreases. Doing so here will cause the _getPayout to break as it always relies on the poolAmount to determine how much each recipient gets. This change ignores reducing the poolAmount but this behaviour doesn't align with the other strategies. Need to discuss

### DIFF
--- a/contracts/strategies/qv-base/QVBaseStrategy.sol
+++ b/contracts/strategies/qv-base/QVBaseStrategy.sol
@@ -97,6 +97,9 @@ abstract contract QVBaseStrategy is BaseStrategy {
     /// @notice Whether or not the strategy requires metadata
     bool public metadataRequired;
 
+    /// @notice Whether the distribution started or not
+    bool public distributionStarted;
+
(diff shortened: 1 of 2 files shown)


0xKurt:
suggestion:

```
    function withdraw(address _token) external onlyPoolManager(msg.sender) {
        if (block.timestamp <= allocationEndTime + 30 days) {
            revert INVALID();
        }

        uint256 amount;

        if(_token == NATIVE)  {
            amount = address(this).balance;
        } else {
            amount = ERC20(token).balanceOf(address(this));
        }
        // Transfer the tokens to the 'msg.sender' (pool manager calling function)
        _transferAmount(_token, msg.sender, amount);
    }
    ```

0xKurt (approved):

0xKurt (approved):

PR #346, opened 2023-10-04
fix: prevent front-running of proposalBid

thelostone-mc:
refs https://github.com/sherlock-audit/2023-09-Gitcoin-judging/issues/497

### DIFF
--- a/contracts/strategies/rfp-simple/RFPSimpleStrategy.sol
+++ b/contracts/strategies/rfp-simple/RFPSimpleStrategy.sol
@@ -412,7 +412,7 @@ contract RFPSimpleStrategy is BaseStrategy, ReentrancyGuard {
             // If the proposal bid is not equal to the final proposal bid this will revert
             // This is to prevent the pool manager from decreasing the proposal bid
             // or recipient from front running and increasing the proposal bid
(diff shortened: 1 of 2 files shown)


contracts/strategies/rfp-simple/RFPSimpleStrategy.sol
```diff
@@ -407,12 +408,19 @@ contract RFPSimpleStrategy is BaseStrategy, ReentrancyGuard {
         // Update status of acceptedRecipientId to accepted
         recipient.recipientStatus = Status.Accepted;
 
+        if (recipient.proposalBid != finalProposalBid) {
+            // If the proposal bid is not equal to the final proposal bid this will revert
+            // This is to prevent the pool manager from decreasing the proposal bid
+            // or recipient from front running and increasing the proposal bid
+            revert AMOUNT_TOO_LOW();
```
0xKurt, line 415: since there could be different reasons of why the proposal bids do not match, i would suggest to use a more generic error like `INVALID`

0xKurt (changes requested):

codenamejason (approved):

PR #349, opened 2023-10-09
fix: update review status

0xKurt:
Refs https://github.com/sherlock-audit/2023-09-Gitcoin-judging/issues/729
Refs https://github.com/sherlock-audit/2023-09-Gitcoin-judging/issues/699


codenamejason (dismissed):

thelostone-mc:
TODO: check if we need to revisit the reviewRecipients logic cause we use a threshold based approval (which is 2)
So if i have 2 approvals and then i re-register, if anyone reviews my -> my approval gets to 3 (since the old approvals are still preserved)

PR #358, opened 2023-10-12
zksync create2

thelostone-mc:


0xKurt:
This pr includes to many changes regarding deploy scripts, test deployments, etc.. 
For now we should keep it for the later updates of the scripts, but for the audit it is a way too much unnecessary stuff
@codenamejason @thelostone-mc

PR #370, opened 2023-10-16
fix: revert on removing allocator after voting

thelostone-mc:
refs https://github.com/sherlock-audit/2023-09-Gitcoin-judging/issues/468


thelostone-mc:
Closing as this is expected behaviour

PR #392, opened 2023-10-18
Add  deploy checklist / update deploy scripts

codenamejason:
Changes Included:
1.  Adds updated deploy scripts.
2. Adds deploy all networks with one command (make sure all wallets are funded or this will fail).
3. Adds gas reports.
4. Adds snapshots.
5. Adds updated to use bun instead of yarn --> happy to chat about this if anyone is not pleased about it but bun is way faster.
6. Adds all fixes for audit from another PR.
7. Adds verification.


thelostone-mc:
Let's figure out a neater way as opposed to documenting every chain and all the steps.
Ideally we'd want something detailed which we can refer for every chain . Kinda like a deploy template (with deploy commands so that it's useful )

https://github.com/allo-protocol/allo-contracts/blob/v1.0.0/docs/DEPLOY_STEPS.md

codenamejason:
I have updated to bun for running command and managing the packages, it is sooo much faster you wont believe it. Here is the install docs to make sure you have globally on your machine --> https://bun.sh/docs/installation

@thelostone-mc @0xKurt

hardhat.config.ts
```diff
@@ -277,41 +370,41 @@ export default config;
...
-    const outputPath = path.resolve(__dirname, "DEPLOYMENT_INFORMATION.md");
-    fs.appendFileSync(outputPath, mdContent.trim());
+//     // Write the formatted information to a Markdown file
+//     const outputPath = path.resolve(__dirname, "DEPLOYMENT_INFORMATION.md");
+//     fs.appendFileSync(outputPath, mdContent.trim());
 
-    console.log(`Information has been written to ${outputPath}`);
-  });
+//     console.log(`Information has been written to ${outputPath}`);
+//   });
```
thelostone-mc, line 410: ^ why are we commenting this out

package.json
```diff
@@ -56,8 +59,8 @@
     "eslint-plugin-no-only-tests": "^3.1.0",
     "eslint-plugin-node": "^11.1.0",
     "ethereum-waffle": "^4.0.10",
-    "ethers": "^6.8.0",
-    "hardhat": "^2.18.2",
+    "ethers": "^6.7.1",
+    "hardhat": "^2.17.4",
```
thelostone-mc, line 63: ^ Do we need to downgrade ?

scripts/core/deployContractFactory.ts
```diff
@@ -47,10 +48,7 @@ export async function deployContractFactory() {
   };
 
   deploymentIo.write(objToWrite);
-
   await delay(20000);
-  await verifyContract(instance.target.toString(), []);
```
thelostone-mc, line 52: Shouldn't we keep the verifying?

foundry.toml
```diff
@@ -7,9 +7,19 @@ src = 'contracts'
...
 remappings = []
 
+# Gas reporting
+gas_reports = ["ContractFactory", "Registry", "Allo"]
+gas_reports_ignore = [
+  "/test/utils/*",
+  # "TestStrategy",
+  # "QVBaseStrategyTestMock",
+  # "Permit2",
+  # "MockStrategy",
```
thelostone-mc, line 20: ^ Shouldn't these be uncommented

package.json
codenamejason: This was left from zk deployment. Will update.

hardhat.config.ts
codenamejason: it's not working and will be removed

package.json
thelostone-mc: @0xKurt just calling this out

thelostone-mc (approved):

0xKurt (approved):

PR #416, opened 2023-11-07
Remove check on register recipient

codenamejason:
Removing this check for Impact Stream Strategy -->

```solidity
[438] if (isUsingRegistryAnchor && !_isProfileMember(recipientId, recipientAddress)) revert UNAUTHORIZED();
```

@0xKurt and I deployed this new version to run the scripts from on Goerli.

cc: @0xKurt @thelostone-mc


codenamejason:
I also think we need to add this:

```solidity
 function getTotalVotesForRecipient(address _recipientId) external view returns (uint256) {
      return recipients[_recipientId].totalVotesReceived;
  }
```

cc: @thelostone-mc @0xKurt

thelostone-mc (approved):

PR #428, opened 2023-11-27
Update naming

codenamejason:


thelostone-mc:
Closing after discussing that we want to retain the microgrants name for contract/spec

PR #439, opened 2023-12-14
Deploy Arb-Sepolia - Remove Arb-Goerli

codenamejason:


scripts/test/createProfile.ts
```diff
@@ -13,7 +12,7 @@ export async function createProfile() {
 
     const balance = await ethers.provider.getBalance(deployerAddress);
 
-    const registry = registryConfig[chainId].registryProxy;
+    const registry = "0x4AAcca72145e1dF2aeC137E1f3C5E3D75DB8b5f3";
```
thelostone-mc, line 15: ^ can we change this back to fetch it from the config ?

.openzeppelin/unknown-421614.json
```diff
@@ -0,0 +1,515 @@
+{
```
thelostone-mc, line 1: we can remove the arb goerli as it's not used

scripts/config/allo.config.ts
```diff
@@ -95,10 +95,10 @@ export const alloConfig: DeployParams = {
...
     baseFee: 0,
   },
-  // Arbitrum Goerli
-  421613: {
-    owner: "0xE64c2171bB7c7412eb2ad3c4c027C499FEf345DC",
-    treasury: "0xE64c2171bB7c7412eb2ad3c4c027C499FEf345DC",
+  // Arbitrum Sepolia
+  421614: {
+    owner: "0x8C180840fcBb90CE8464B4eCd12ab0f840c6647C",
+    treasury: "0x8C180840fcBb90CE8464B4eCd12ab0f840c6647C",
```
thelostone-mc, line 101: Is this a reason we use a different EOA than we what had used on Arb Goeri ?

thelostone-mc (changes requested):
Left in a few comments but otherwise LGTM

scripts/test/createProfile.ts
codenamejason: This doesn't exist on the config currently. We never updated the script when we updated the configs.

Two Options:
1. Just hardcode when we test creating a profile and pool after deploy on testnets.
2. Add the `registryProxy` to the config on all networks (has to be added to the type).

.openzeppelin/unknown-421614.json
codenamejason: I am curious to why the `arbitrum-sepolia.json` file is not present 🤔

scripts/config/allo.config.ts
codenamejason: There were a few different ones so we used the one with funds and was already present in the code.

thelostone-mc (approved):

PR #450, opened 2024-01-03
Error reproduction. Import Hats. 

jordanlesich:
When I add this import statement to test/foundry/strategies/DirectGrantsSimpleStrategy.t.sol

I get this error:

```
Failed to resolve file: "XXXXXXXX/repos/allo-hats-issue/node_modules/solady/utils/Multicallable.sol": No such file or directory (os error 2).
 Check configured remappings..
    --> "XXXXXXX/repos/allo-hats-issue/lib/hats-protocol/src/Hats.sol"
        "solady/utils/Multicallable.sol"
```
        

the libs directive in foundry.toml causes the remapping to look in node_modules first. 

image

I was able to create a temporary fix by switching ``libs = ['node_modules', 'lib']`` to ``libs = ['lib', 'node_modules']`` **and** going into the Hats submodule and changing the import statement for multicallable to a relative import instead of a remapping.

 
image

For obvious reasons, making a change to submodule is a temporary fix. If I wish to push my changes anywhere, I'll need a much better fix. 

A few things worth noting. 

- Changing the remappings for solady at the root level creates incorrect import paths. 
- The remappings to solady within Hats seem to be correct. 
- Tests pass if I use the temporary fix.


test/foundry/strategies/DirectGrantsSimpleStrategy.t.sol
```diff
@@ -14,6 +14,7 @@ import {AlloSetup} from "../shared/AlloSetup.sol";
 import {RegistrySetupFull} from "../shared/RegistrySetup.sol";
 
 import {EventSetup} from "../shared/EventSetup.sol";
+import {Hats} from "hats-protocol/Hats.sol";
```
thelostone-mc, line 17: TODO: remove this before merging if we are alright with the reampping update @0xKurt @codenamejason

thelostone-mc:
```suggestion
```

thelostone-mc (approved):

PR #472, opened 2024-02-16
Permit and DAI permit support

gravityblast:
closes https://github.com/gitcoinco/grants-stack/issues/2937


contracts/strategies/donation-voting-merkle-distribution-direct-transfer/DonationVotingMerkleDistributionDirectTransferStrategy.sol
```diff
@@ -70,9 +72,48 @@ contract DonationVotingMerkleDistributionDirectTransferStrategy is DonationVotin
...
+            catch Error(string memory reason) {
+                if (IERC20(token).allowance(msg.sender, address(this)) < amount) {
+                    revert(reason);
+                }
+            } catch (bytes memory reason) {
+                if (IERC20(token).allowance(msg.sender, address(this)) < amount) {
+                    revert(string(reason));
+                }
+            }
+            IERC20(token).transferFrom(_sender, _recipients[recipientId].recipientAddress, amount);
```
0xKurt, line 89: On etherscan the transaction will be marked as Failed if the catch is entered. this might cause a lot of confusion.

wdyt about implementing it like that:

```
IERC20 t = IERC20(token);
if (t.allowance(msg.sender, address(this)) < amount) {
    t.permit(_sender, address(this), amount, p2Data.permit.deadline, v, r, s);
}
t.transferFrom(_sender, _recipients[recipientId].recipientAddress, amount);
 ```

Comment (gravityblast) [reply contracts/strategies/donation-voting-merkle-distribution-direct-transfer/DonationVotingMerkleDistributionDirectTransferStrategy.sol:89]: great point @0xKurt 

we have 2 options. 

1 - the current implementation is checking the allowance only when the call to `permit` fails, and it can fail only in very rare cases when the signed permit expired. It can happen front running the tx or using the same nonce before the tx is mined. 

2 - the other option you proposed is to always check the allowance and call `permit` only when it's needed. 

Given that this problem is so rare, I would like to go with option 1, so that we never use more gas then needed calling the other contract,  unless the very rare case happens.

**Review by 0xKurt — APPROVED** [review]

File: contracts/strategies/donation-voting-merkle-distribution-direct-transfer/DonationVotingMerkleDistributionDirectTransferStrategy.sol (hunk 1 above)
Comment (0xKurt) [reply contracts/strategies/donation-voting-merkle-distribution-direct-transfer/DonationVotingMerkleDistributionDirectTransferStrategy.sol:89]: lets keep the  current implementation

**Review by 0xKurt — DISMISSED** [review]

File: contracts/strategies/donation-voting-merkle-distribution-direct-transfer/DonationVotingMerkleDistributionDirectTransferStrategy.sol [hunk 2]
```diff
@@ -44,7 +47,8 @@ contract DonationVotingMerkleDistributionDirectTransferStrategy is DonationVotin
     /// @param _sender The sender of the allocation
     function _afterAllocate(bytes memory _data, address _sender) internal override {
```
codenamejason, line 48: Can we update the `_data` param to reflect the changes.

contracts/strategies/donation-voting-merkle-base/DonationVotingMerkleDistributionBaseStrategy.sol
```diff
@@ -663,7 +668,7 @@ abstract contract DonationVotingMerkleDistributionBaseStrategy is Native, BaseSt
     /// @param _sender The sender of the transaction
     function _allocate(bytes memory _data, address _sender) internal virtual override onlyActiveAllocation {
```
codenamejason, line 669: Also update param `_data` here.

thelostone-mc (approved):

thelostone-mc (approved):

PR #474, opened 2024-02-19
SQF: Emergency Functions, Super App factory

codynhat:
Adds two emergency functions to the recipient Super Apps in the SQF Superfluid strategy to allow recipients to withdraw funds and close incoming streams.

Creates a factory to deploy the recipient Super Apps since the above change put the contract size over the deploy limit.

Also makes the Super App registration key configurable upon deployment to support deploying to mainnet.


codynhat:
@thelostone-mc Thanks. Just ran it.

thelostone-mc (approved):

PR #478, opened 2024-02-23
add registration start/end time to direct grants strategy

gravityblast:
closes https://github.com/gitcoinco/grants-stack/issues/2961


codenamejason (approved):
Strange to me we are allowing start dates in the past (I'm sure there is a reason). And... I just saw the other PR 😂

thelostone-mc (approved):

0xKurt (changes requested):
Timestamps are unused.

Fix: introduce an `onlyActiveRegistration` modifier which checks the block timestamp and add it to `_registerRecipient`

@gravityblast

gravityblast:
@0xKurt great catch thank you! I pushed a fix

gravityblast:
> Strange to me we are allowing start dates in the past (I'm sure there is a reason). And... I just saw the other PR 😂

the reason is that we need to allow any change in the dates. But if we don't allow to set the start time in the past, we would need to update it to `now` every time we want to extend the end date

0xKurt (approved):

PR #534, opened 2024-04-08
deploy strategy 

thelostone-mc:
### Description

image

### DIFF
--- a/deploy/zksync/strategies/deployEraStrategies.ts
+++ b/deploy/zksync/strategies/deployEraStrategies.ts
@@ -13,17 +13,15 @@ const deployContractUsingFactoryWithBytecodeOnEra = async (
   constructorArgs?: Args,
 ): Promise<string> => {
 
-  // const provider = new Provider("https://sepolia.era.zksync.dev");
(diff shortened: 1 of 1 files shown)


deploy/strategies/deployEraStrategies.ts
```diff
@@ -13,17 +13,15 @@ const deployContractUsingFactoryWithBytecodeOnEra = async (
...
-  // const deployerContract = new ethers.Contract(
-  //   contractFactoryAddress,
-  //   ContractArtifact.abi,
-  //   deployerAddress
-  // );
+  const deployerContract = new ethers.Contract(
+    contractFactoryAddress,
+    ContractArtifact.abi,
+    deployerAddress
+  );
```
thelostone-mc, line 24: ^ this fails with `TypeError: Cannot read properties of undefined (reading 'Contract')`

kiriyaga-txfusion: Hello, we checked it out, and it seems that you want to use `ethers` from `hardhat` but you didn't imported it in the `era.hardhat.config` file. 
```javascript
import "@nomicfoundation/hardhat-ethers"
```
Also, I would recommend to use our plugins (and SDK) in that script for communicating with the zkSync networks.

PR #575, opened 2024-06-07
feat: remove strategy approval

### LINKED ISSUE #574: Remove the need for strategies to be approved
Following on Kevin’s [list](https://github.com/allo-protocol/allo-v2/issues/544) we want to 
> remove the paternalizing requirement that allo team "approves" every strategy.

Context
---
Currently there are 2 options to create a new Round.
1.  With cloneable strategy. Allo stores a mapping with addresses of Strategies that are approved by the protocol and can be safely cloned for a new Round.
2. With a custom strategy. The user provides an already deployed Strategy contract that is not part of the accepted cloneable mapping. 

Solution
---

To achieve that we should update the `Allo` contract as:
- [ ] We assume all strategies cloneable/approved
- [ ] Removal of `cloneableStrategies` mapping
- [ ] Removal of `createPoolWithCustomStrategy` function
- [ ] Removal of `addToCloneableStrategies` function 
- [ ] Removal of `removeFromCloneableStrategies` function
- [ ] Removal of `_isCloneableStrategy` & `isCloneableStrategy` functions
- [ ] Update `createPool`
- [ ] Update docs
- [ ] Update tests

Questions
---
Is this change gonna affect the SDK or any other offchain tool that we should account for?

0xOneTony:
Closes #574 

Decided to keep the `createPoolWithCustomStrategy` instead of completely removing because if someone creates a custom strategy they shouldnt need to deploy and then clone their custom contract, they should just deploy and if someone else wants to use their contract they can by calling `createPool`.

- I run `yarn fmt` so some files that were affected but didnt change the logic were: `MockERC20PermitDAI`, `PermitSigUtils`, `PermitSigUtilsDAI`
- Also updated gas report and table


contracts/core/Allo.sol
```diff
@@ -79,10 +79,6 @@ contract Allo is
     /// @dev 'Pool.id' -> 'Pool'
     mapping(uint256 => Pool) private pools;
 
-    /// @notice Returns a bool for whether a strategy is cloneable or not using the strategy address as the key
-    /// @dev Strategy.address -> bool
-    mapping(address => bool) private cloneableStrategies;
```
0xKurt, line 84: Removing a mapping disrupts this consistency of the storage layout and could lead to unexpected behavior if we need to add new storage variables in a future upgrade.
Consider keeping it with a comment marking it as deprecated/not used.

thelostone-mc (approved):

PR #604, opened 2024-06-26
feat: add functions to add and remove multiple managers

### LINKED ISSUE #581: Accept multiple addresses on Allo.sol
One way of achieving this would be to update addProfileManager and removeProgramManager.

### LINKED ISSUE #585: Accept multiple Pool Managers
One way of achieving this would be to update addPoolManager and removePoolManager.

0xAustrian:
Closes #585

- [x] Create `addPoolManagers` and `removePoolManagers` functions.
- [x] Add the new functions to the interface.
- [x] Update `gas-table` and `gas-report`.
- [x] Update README.
- [x] Add tests for the new functions.
- [x] Remove `addPoolManager` and `removePoolManager` (and it's tests, interfaces, etc).


contracts/core/Allo.sol
```diff
@@ -255,6 +255,22 @@ contract Allo is
...
+    function addPoolManagers(uint256 _poolId, address[] memory _managers) external onlyPoolAdmin(_poolId) {
+        for (uint256 i; i < _managers.length;) {
+            if (_managers[i] == address(0)) revert ZERO_ADDRESS();
+
+            _grantRole(pools[_poolId].managerRole, _managers[i]);
+
+            unchecked {
+                ++i;
+            }
+        }
```
0xOneTony, line 271:
```suggestion
        for (uint256 i; i < _managers.length;) {
            addPoolManager(_poolId, manager[i]);

            unchecked {
                ++i;
            }
        }
```

I believe we can also make `addPoolManager` internal and remove the `onlyPoolAdmin` modifier since it is being called on `addPoolManagers`

contracts/core/Allo.sol
```diff
@@ -263,6 +279,20 @@ contract Allo is
         _revokeRole(pools[_poolId].managerRole, _manager);
     }
 
+    /// @notice Remove multiple pool managers
+    /// @dev Emits 'RoleRevoked()' event. 'msg.sender' must be a pool admin.
+    /// @param _poolId ID of the pool
+    /// @param _managers The addresses to remove
+    function removePoolManagers(uint256 _poolId, address[] memory _managers) external onlyPoolAdmin(_poolId) {
```
0xOneTony, line 286: Same thing with `addPoolManagers`

0xOneTony (commented):
Nice! Mentioned some small improvements

contracts/core/Allo.sol
```diff
@@ -255,6 +255,22 @@ contract Allo is
         _grantRole(pools[_poolId].managerRole, _manager);
     }
 
+    /// @notice Add multiple pool managers
+    /// @dev Emits 'RoleGranted()' event. 'msg.sender' must be a pool admin.
+    /// @param _poolId ID of the pool
+    /// @param _managers The addresses to add
+    function addPoolManagers(uint256 _poolId, address[] memory _managers) external onlyPoolAdmin(_poolId) {
```
ilpepepig, line 262: I think `calldata` is more appropriate than `memory` here. `_managers` won't get modified and using `calldata` is slightly more efficient.

contracts/core/Allo.sol
ilpepepig, line 286: Same `calldata` comment as in `addPoolManagers`

0xAustrian: I did it only with `addPoolManagers` since `removePoolManagers` is only one line and it will not make difference.

contracts/core/Allo.sol
```diff
@@ -594,6 +602,18 @@ contract Allo is
         emit BaseFeeUpdated(baseFee);
     }
 
+    /// @notice Adds a pool manager
+    /// @dev Internal function used to add a pool manager.
+    /// @param _poolId The ID of the pool
+    /// @param _manager The address to add
+    function _addPoolManager(uint256 _poolId, address _manager) internal {
```
ilpepepig, line 609: Now that we have this function maybe we should use `_addPoolManager` for https://github.com/allo-protocol/allo-v2/blob/87f18bbad4302f1c28d7ad669bddad301c499503/contracts/core/Allo.sol#L458-L461

0xOneTony: true! nice notice

0xAustrian: nice catch!

ilpepepig (approved):

0xOneTony (approved):
LGTM! I believe it makes sense to support multiple Pool Managers since they have more limited authority over Pool Admins. One example is QVBaseStrategy and its derived contracts that support a review function which is guarded by the `onlyPoolManager` modifier, leveraging multiple Pool Managers in this case is beneficial.

contracts/core/Allo.sol
```diff
@@ -243,24 +243,32 @@ contract Allo is
...
     /// @dev Emits 'RoleGranted()' event. 'msg.sender' must be a pool admin.
     /// @param _poolId ID of the pool
-    /// @param _manager The address to add
-    function addPoolManager(uint256 _poolId, address _manager) external onlyPoolAdmin(_poolId) {
-        // Reverts if the address is the zero address with 'ZERO_ADDRESS()'
-        if (_manager == address(0)) revert ZERO_ADDRESS();
+    /// @param _managers The addresses to add
+    function addPoolManagers(uint256 _poolId, address[] calldata _managers) external onlyPoolAdmin(_poolId) {
+        for (uint256 i; i < _managers.length;) {
+            _addPoolManager(_poolId, _managers[i]);
```
thelostone-mc, line 252: Do we need the internal function ? All we have is a zero check which is an extra check than what is on removePoolManagers
Might be worth keeping it consistent with removePoolManagers.

image

0xAustrian: The internal is made to avoid repeating code ([used in _createPool](https://github.com/allo-protocol/allo-v2/blob/feat/multiple-managers/contracts/core/Allo.sol#L459)). And `removePoolManagers` does not have an internal because is only one line. I can create `_removePoolManager` for consistency if needed.

PR #605, opened 2024-06-27
feat: implement multiple profile owners

### LINKED ISSUE #582: Add or remove multiple profile owners
Evaluate if it makes sense to add and remove multiple profile owners or just expect folks to use safe

0xAustrian:
Closes #582

- [x] Modify `_owner` param in `createProfile()`.
- [x] Use same Role system as members for owners.
- [x] Delete `updateProfilePendingOwner` and `acceptProfileOwnership` functions.
- [x] Delete `profileIdToPendingOwner` mapping.
- [x] Add `addOwners` and `removeOwners` functions.
- [x] Update interfaces.
- [x] Update tests suite.
- [x] Add tests for new functions.
- [x] Delete unused code.
- [x] Temporarily disabled `SQFSuperFluidStrategyTest` E2E (added a TODO).
- [x] Update README.
- [x] Update `gas-report` and `gas-table`.


contracts/migration/AlloV1ToV2ProfileMigration.sol
```diff
@@ -84,23 +84,25 @@ contract AlloV1ToV2ProfileMigration {
         bytes32[] memory migratedProfileIds = new bytes32[](_profilesLength);
 
         for (uint256 i = 0; i < _profilesLength;) {
-            bytes32 alloV1 = _projectIds[i];
-            uint256 alloV1ChainId = _sourceChainIds[i];
-            uint256 nonce = _nonces[i];
```
0xAustrian, line 89: Had to remove those 3 because of `stack too deep`

contracts/core/Registry.sol
```diff
@@ -112,61 +108,69 @@ contract Registry is IRegistry, Initializable, Native, AccessControlUpgradeable,
...
         profilesById[profileId] = profile;
         anchorToProfileId[profile.anchor] = profileId;
 
-        // Assign roles for the profile members
-        uint256 memberLength = _members.length;
+        bytes32 PROFILE_MEMBER_ROLE = profileId;
+        bytes32 PROFILE_OWNER_ROLE = keccak256(abi.encodePacked(profileId, "owner"));
+
+        // Assign roles for the profile owners
+        _grantRole(PROFILE_OWNER_ROLE, msg.sender);
```
0xOneTony, line 143: Why do we give the owner role to the msg.sender?

contracts/core/Registry.sol
```diff
@@ -112,61 +108,69 @@ contract Registry is IRegistry, Initializable, Native, AccessControlUpgradeable,
...
+            address owner = _owners[i];
+
+            // Will revert if any of the addresses are a zero address
+            if (owner == address(0)) revert ZERO_ADDRESS();
+
+            // Grant the role to the owner and emit the event for each owner
+            _grantRole(PROFILE_OWNER_ROLE, owner);
+            unchecked {
+                ++i;
+            }
```
0xOneTony, line 155: We could use the `addOwners` or `_addOwner` here

contracts/core/Registry.sol
0xAustrian: For simplicity of the code, and if you provide an empty array of owners, no error is thrown and you can have only one owner.

contracts/core/Registry.sol
```diff
@@ -49,10 +49,6 @@ contract Registry is IRegistry, Initializable, Native, AccessControlUpgradeable,
     /// @dev Profile.id -> Profile
     mapping(bytes32 => Profile) public profilesById;
 
-    /// @notice This maps the profile ID to the pending owner
-    /// @dev Profile.id -> pending owner
-    mapping(bytes32 => address) public profileIdToPendingOwner;
```
ilpepepig, line 54: Being this an upgradeable contract, should we remove the `profileIdToPendingOwner` declaration? It would be easier to review future upgrades if we leave it as private and change its name to something like `profileIdToPendingOwner_deprecated`.

contracts/core/Registry.sol
```diff
@@ -321,6 +284,37 @@ contract Registry is IRegistry, Initializable, Native, AccessControlUpgradeable,
         }
     }
 
+    /// @notice Adds owners to the profile
+    /// @dev 'msg.sender' must be the owner of the profile.
+    /// @param _profileId The ID of the profile
+    /// @param _owners The owners to add
+    function addOwners(bytes32 _profileId, address[] memory _owners) external onlyProfileOwner(_profileId) {
```
ilpepepig, line 291: Owners are not tracked through events. Should they?

thelostone-mc:
So quick callout here!
We had to implement the complicated transfer and accept ownership to try and prevent me from creating a profile where I'm a member and then transfer it to someone from wonderland. 
Then i could use this to apply To Rounds and get funds to my wallet on your behalf 

@0xAustrian @0xOneTony @ilpepepig Curious on what you're thoughts here are ?

Issue #594, opened 2024-06-11
Ensure core contracts support offchain voting/ meta-transactions


thelostone-mc:

0xOneTony:
Can you explain a bit more the scope of the task?

thelostone-mc:
The scope here is that we have clients who would would want to do all the allocation off-chain and use a relayer to pay fees. We'd just have to validate that the core protocol supports a relayer to pay the fees / enable a strategy can be written to support this

PR #611, opened 2024-07-03
feat: improve msg.sender and msg.value usage

### LINKED ISSUE #593: Improve usage of msg.value and msg.sender
msg.value replaced with _msgValue + msg.sender replaced with _msgSender

ilpepepig:
Closes #593.

- [x] Move msg.sender usages out of internal functions.
- [x] Use OZ:Context::_msgSender() instead of msg.sender.
- [x] Move msg.value usages out of internal functions.
- [x] Update interface.

Supporting meta-transactions as suggested in #594 would probably become as easy as overriding `_msgSender()`. Depending on how meta-transactions are expected to work in Allo, [OZ::ERC2771Context.sol](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/metatx/ERC2771Context.sol) could be used.

The usage of `onlyOwner` remains unchanged for the following reasons:
1. Using a trusted forwarder contract to support meta-transactions adds a risk layer to Allo. For users this seems reasonable, but for such a critical role as the owner of the protocol it seems like an unnecessary risk.
2. The owner address will probably be a contract which can't sign transactions anyway.
3. Owner transactions will probably take place with much less frequency than users transactions.


contracts/core/Allo.sol
0xAustrian, line 1: Remember to update natspec of the functions, I still see some `msg.sender` in `createPoolWithCustomStrategy()`, `createPool()`, `_allocate()`, `_createPool()`

0xAustrian (commented):
Just one small comment.

0xOneTony (approved):
LGTM!

0xAustrian (approved):

thelostone-mc (approved):
Changes LGTM but noob question.
If a client wanted to use this, to pay for their users fees 
I'm assuming if a client wanted to use meta txns for it's users and pay fees

We would need a function which does https://github.com/bcnmy/metatx-standard/blob/master/src/contracts/EIP712MetaTransaction.sol#L37
So that any relayer can sign the msg -> invoke this function and
- get the real msg.sender
- override _msgSender
- know which function to call with what data

Is this right ? (I'm assuming we'd do this in a follow up ticket)

0xOneTony:
> We would need a function which does https://github.com/bcnmy/metatx-standard/blob/master/src/contracts/EIP712MetaTransaction.sol#L37 So that any relayer can sign the msg -> invoke this function and
> 
> * get the real msg.sender
> * override _msgSender
> * know which function to call with what data
> 
> Is this right ? (I'm assuming we'd do this in a follow up ticket)

@thelostone-mc Correct! This will be part of https://github.com/allo-protocol/allo-v2/issues/594

PR #614, opened 2024-07-05
feat: QF helper library

### LINKED ISSUE #609: Add QFLibrary for Quadratic Funding
## Problem
Basic mechanisms like Quadratic Voting/Funding should be abstracted into a trusted, optimized library. Provide helper functions for specific actions in different strategies.

## Solution
- Implement the logic for voting and calculating matching amount for each recipient with Quadratic Funding formula.
- Unit Tests

### Resources
Spec for QFLibrary

```solidity
library QFHelper {
    using EnumerableSet for EnumerableSet.AddressSet;
    using SafeMath for uint256;

    /// Struct that defines a donation
    struct Donation {
        uint256 amount;
        address funder;
    }

    /// Struct that defines the state of the donations to recipients
    struct State {
        EnumerableSet.AddressSet recipients;
        mapping(address => Donation[]) donations;
        /// The square of total contributions for each recipient
        mapping(address => uint256) contributionsSquare;
    }

    /// Calculate the square root of a given number
    function _sqrt(uint256 x) internal pure returns (uint256 y) {}

    /// Donate to a recipient to allocate
    function fundRecipients(
        State storage _state,
        address[] memory _recipients,
        uint256[] memory _amounts
    ) internal {}

    function getPayOutForRecipient(State storage _state, address _recipient) internal view returns (uint256) {}

    function calculateDivisor(uint256 _matchingAmount, uint256 _totalContributions) internal view returns (uint256) {
        return _matchingAmount.div(_totalContributions);
    }

    /// Quadratic Formula
    function calculatePayout(State storage _state, address _recipient, uint256 _matchingAmount) internal view returns (uint256 _recipientPayout) {}
}
```

0xOneTony:
Closes #609 
Quadratic Funding Library that can be used by different strategies to handle voting and calculating the matching amount.

### DIFF
--- a/contracts/core/libraries/QFHelper.sol
+++ b/contracts/core/libraries/QFHelper.sol
@@ -1,7 +1,6 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity 0.8.19;
 
-import "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
 import "solady/utils/FixedPointMathLib.sol";
 
 /// @title QF Helper Library
@@ -9,9 +8,6 @@ import "solady/utils/FixedPointMathLib.sol";
 /// @dev Handles the donation and voting of recipients and calculates the matching amount
 ///      for each recipient using the Quadratic Funding formula
 library QFHelper {
-    /// Using EnumerableSet for EnumerableSet.AddressSet to store the recipients
-    using EnumerableSet for EnumerableSet.AddressSet;
-
     /// @notice Error thrown when the number of recipients and amounts are not equal
     error QFHelper_LengthMissmatch();
 
@@ -24,24 +20,24 @@ library QFHelper {
     }
 
     /// Struct that defines the state of the donations to recipients
-    /// @param recipients The set of recipients
     /// @param sqrtDonationsSum The sum of the square root of the donations for each recipient
     /// @param donations The donations for each recipient
     /// @param totalContributions The total contributions of all recipients
     struct State {
-        EnumerableSet.AddressSet recipients;
         mapping(address => uint256) sqrtDonationsSum;
         mapping(address => Donation[]) donations;
         uint256 totalContributions;
     }
 
     /// @notice Votes for recipients by donating
+    /// @dev The number of recipients and amounts should be equal and the same index
+    ///      should correspond to the same recipient and amount
+    /// @dev We also calculate the total contributions and sqrt sum of donations for each recipient
+    ///      which will be needed to calculate the matching amount
     /// @param _state The state of the donations
     /// @param _recipients The recipients to donate to
     /// @param _amounts The amounts to donate to each recipient
     /// @param _funder The address of the funder
-    /// @dev The number of recipients and amounts should be equal and the same index
-    ///      should correspond to the same recipient and amount
     function fund(State storage _state, address[] memory _recipients, uint256[] memory _amounts, address _funder)
         internal
     {
@@ -49,34 +45,21 @@ library QFHelper {
         /// Check if the number of recipients and amounts are equal
         if (_recipientsLength != _amounts.length) revert QFHelper_LengthMissmatch();
 
+        uint256 _totalContributionsDelta;
         for (uint256 i = 0; i < _recipientsLength; i++) {
-            /// Add the recipient to the set if it doesn't exist
-            if (!_state.recipients.contains(_recipients[i])) {
-                _state.recipients.add(_recipients[i]);
-            }
             /// Add the donation to the recipient
             _state.donations[_recipients[i]].push(Donation({amount: _amounts[i], funder: _funder}));
 
             /// Calculate the square root of the donation amount and add it to the sum of donations
             uint256 _sqrtDonationsSum = _state.sqrtDonationsSum[_recipients[i]];
             _sqrtDonationsSum += FixedPointMathLib.sqrt(_amounts[i]);
+
+            /// Calculate the total contributions delta
+            _totalContributionsDelta += _sqrtDonationsSum ** 2 - _state.sqrtDonationsSum[_recipients[i]] ** 2;
+
             _state.sqrtDonationsSum[_recipients[i]] = _sqrtDonationsSum;
         }
-    }
-
-    /// @notice Calculates and stores the total contributions of all recipients
-    /// @dev The total contributions is the sum of the square of the square root of the donations
-    ///      for each recipient. This should only be called once after all donations have been made
-    /// @param _state The state of the donations
-    function calculateTotalContributions(State storage _state) internal {
-        uint256 _totalContributions;
-        for (uint256 i = 0; i < _state.recipients.length(); i++) {
-            address _recipient = _state.recipients.at(i);
-            uint256 _sqrtDonationsSum = _state.sqrtDonationsSum[_recipient];
-            _totalContributions += _sqrtDonationsSum * _sqrtDonationsSum;
-        }
-
-        _state.totalContributions = _totalContributions;
+        _state.totalContributions += _totalContributionsDelta;
     }
 
     /// @notice Calculates the matching amount for a recipient using the Quadratic Funding formula
--- a/contracts/core/libraries/QFHelper.sol
+++ b/contracts/core/libraries/QFHelper.sol
@@ -1,7 +1,6 @@
 // SPDX-License-Identifier: UNLICENSED
-pragma solidity ^0.8.19;
+pragma solidity 0.8.19;
 
-import "@openzeppelin/contracts/utils/math/SafeMath.sol";
 import "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
 
 /// @title QF Helper Library
@@ -11,8 +10,6 @@ import "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
 library QFHelper {
     /// Using EnumerableSet for EnumerableSet.AddressSet to store the recipients
     using EnumerableSet for EnumerableSet.AddressSet;
-    /// Using SafeMath for uint256 calculations
-    using SafeMath for uint256;
 
     /// @notice Error thrown when the number of recipients and amounts are not equal
     error QFHelper_LengthMissmatch();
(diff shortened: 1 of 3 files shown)


contracts/core/libraries/QFHelper.sol
```diff
@@ -0,0 +1,120 @@
...
+
+    /// @notice Calculates the matching amount for each recipient using the Quadratic Funding formula
+    /// @param _state The state of the donations
+    /// @param _matchingAmount The total matching amount
+    /// @return _recipients The recipients
+    /// @return _payouts The matching amount for each recipient
+    function calculateMatching(State storage _state, uint256 _matchingAmount)
+        internal
+        view
+        returns (address[] memory _recipients, uint256[] memory _payouts)
```
carlbarrdahl, line 76: Could we rename payouts to amounts for consistency?

test/utils/MockQFHelper.sol
```diff
@@ -0,0 +1,28 @@
...
+pragma solidity ^0.8.19;
+
+import "contracts/core/libraries/QFHelper.sol";
+
+/// @title Mock QF Helper
+/// @notice A mock contract for testing Quadratic Funding Library
+contract MockQFHelper {
+    using QFHelper for QFHelper.State;
+
+    QFHelper.State internal state;
```
carlbarrdahl, line 11: This is super cool way to use libraries!

test/utils/MockQFHelper.sol
```diff
@@ -0,0 +1,28 @@
...
+
+    function fundRecipients(address[] memory _recipients, uint256[] memory _amounts) public {
+        state.fundRecipients(_recipients, _amounts);
```
carlbarrdahl, line 14: any chance we could call it just `state.fund(_recipients, _amounts)`? to me it's redundant with `fundRecipients(_recipients, ...`

contracts/core/libraries/QFHelper.sol
```diff
@@ -0,0 +1,120 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.19;
+
+import "@openzeppelin/contracts/utils/math/SafeMath.sol";
```
0xAustrian, line 4: Why is is needed?

contracts/core/libraries/QFHelper.sol
```diff
@@ -0,0 +1,120 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.19;
```
0xAustrian, line 2: In other files the version is fixed.

contracts/core/libraries/QFHelper.sol
```diff
@@ -0,0 +1,120 @@
...
+    /// @param donations The donations for each recipient
+    struct State {
+        EnumerableSet.AddressSet recipients;
+        mapping(address => Donation[]) donations;
+    }
+
+    /// @notice Calculate the square root of a number (Babylonian method)
+    /// @param x The number
+    /// @return y The square root
+    function _sqrt(uint256 x) internal pure returns (uint256 y) {
```
0xAustrian, line 39: To be consistent with other functions in the file parameters should be `_x` and `_y`.

contracts/core/libraries/QFHelper.sol
```diff
@@ -0,0 +1,120 @@
...
+        /// Check if the number of recipients and amounts are equal
+        if (_recipients.length != _amounts.length) revert QFHelper_LengthMissmatch();
+
+        for (uint256 i = 0; i < _recipients.length; i++) {
+            /// Add the recipient to the set if it doesn't exist
+            if (!_state.recipients.contains(_recipients[i])) {
+                _state.recipients.add(_recipients[i]);
+            }
+            /// Add the donation to the recipient
+            _state.donations[_recipients[i]].push(Donation({amount: _amounts[i], funder: msg.sender}));
```
0xAustrian, line 64: Should we send the funder as a parameter here?

contracts/core/libraries/QFHelper.sol
```diff
@@ -0,0 +1,120 @@
...
+        returns (address[] memory _recipients, uint256[] memory _amounts)
+    {
+        /// Get the number of recipients
+        uint256 _numRecipients = _state.recipients.length();
+        /// Initialize the arrays
+        _recipients = new address[](_numRecipients);
+        _amounts = new uint256[](_numRecipients);
+
+        uint256[] memory _donationsSum = new uint256[](_numRecipients);
+        uint256 _totalContributions = 0;
```
0xAustrian, line 85: I think this initialization is not needed.

0xAustrian (approved):

contracts/core/libraries/QFHelper.sol
```diff
@@ -0,0 +1,120 @@
...
+        uint256 _numRecipients = _state.recipients.length();
+        /// Initialize the arrays
+        _recipients = new address[](_numRecipients);
+        _amounts = new uint256[](_numRecipients);
+
+        uint256[] memory _donationsSum = new uint256[](_numRecipients);
+        uint256 _totalContributions;
+        uint256 _sumOfSquareRoots;
+        /// Calculate the matching amount for each recipient
+        for (uint256 i = 0; i < _numRecipients; i++) {
```
0xKurt, line 88: If the number of recipients is too high, iterating through all of them could become very expensive and end in a DOS condition

contracts/core/libraries/QFHelper.sol
```diff
@@ -0,0 +1,120 @@
...
+    /// @param _funder The address of the funder
+    /// @dev The number of recipients and amounts should be equal and the same index
+    ///      should correspond to the same recipient and amount
+    function fund(State storage _state, address[] memory _recipients, uint256[] memory _amounts, address _funder)
+        internal
+    {
+        /// Check if the number of recipients and amounts are equal
+        if (_recipients.length != _amounts.length) revert QFHelper_LengthMissmatch();
+
+        for (uint256 i = 0; i < _recipients.length; i++) {
```
0xKurt, line 58: length can be cached

contracts/core/libraries/QFHelper.sol
```diff
@@ -0,0 +1,120 @@
...
+        for (uint256 i = 0; i < _numRecipients; i++) {
+            /// Get the recipient
+            address recipient = _state.recipients.at(i);
+            /// Set the recipient in the array
+            _recipients[i] = recipient;
+            /// Get the donations for the recipient
+            Donation[] memory _donations = _state.donations[recipient];
+            /// Calculate the sum of the square roots of the donations
+            _sumOfSquareRoots = 0;
+            for (uint256 j = 0; j < _donations.length; j++) {
```
0xKurt, line 97: length could be cached

contracts/core/libraries/QFHelper.sol
0xKurt: i would say yes, since the calculations are just to expensive. I looked into it in the past and couldn't find any good solution.
Maybe there is a third party which we can utilize, like chainlink?

thelostone-mc: So this was a 2 rounds ago https://gov.gitcoin.co/t/gg19-program-round-results/17281/6
It is very realistic to have 1k + unique donors who would have allocated to 1 or more projects.

I do agree with kurt that we would hit our limit pretty quick.
That's why we originally had to do it off-chain.
Additonally for Gitcoin , we end up using passport off-chain service to get scores to assign weights to matching (all which cannot be done on-chain)

At gitcoin would need the off-chain calculation. Even if we keep it as a helper for other to use , they might run it issues like running into gas limit and that feels clunky 

Waddya think @0xOneTony ?

contracts/core/libraries/QFHelper.sol
```diff
@@ -0,0 +1,84 @@
...
+    {
+        /// get the sqrt sum of donations for the recipient
+        uint256 _sqrtDonationsSum = _state.sqrtDonationsSum[_recipient];
+        /// square the sqrt sum of donations
+        uint256 _squareDonationsSum = _sqrtDonationsSum * _sqrtDonationsSum;
+
+        /// calculate the divisor
+        uint256 _divisor = _matchingAmount / _state.totalContributions;
+        /// calculate the matching amount
+        _amount = _squareDonationsSum * _divisor;
```
ilpepepig, line 82: To reduce the loss of precision and to remove the risk of `_matchingAmount` being smaller than `_state.totalContributions` I suggest:
```suggestion
        /// calculate the matching amount
        _amount = _squareDonationsSum * _matchingAmount / _state.totalContributions;
```

contracts/core/libraries/QFHelper.sol
```diff
@@ -0,0 +1,84 @@
...
+        internal
+    {
+        uint256 _recipientsLength = _recipients.length;
+        /// Check if the number of recipients and amounts are equal
+        if (_recipientsLength != _amounts.length) revert QFHelper_LengthMissmatch();
+
+        uint256 _totalContributionsDelta;
+        for (uint256 i = 0; i < _recipientsLength; i++) {
+            /// Add the donation to the recipient
+            _state.donations[_recipients[i]].push(Donation({amount: _amounts[i], funder: _funder}));
```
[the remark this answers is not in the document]
0xOneTony: Right, nice catch now that we do the calculation in fund there is no need for this in the library, the user can save that on-chain if they want on the strategy contract.

contracts/core/libraries/QFHelper.sol
```diff
@@ -0,0 +1,69 @@
...
+
+    /// @notice Votes for recipients by donating
+    /// @dev The number of recipients and amounts should be equal and the same index
+    ///      should correspond to the same recipient and amount
+    /// @dev We also calculate the total contributions and sqrt sum of donations for each recipient
+    ///      which will be needed to calculate the matching amount
+    /// @param _state The state of the donations
+    /// @param _recipients The recipients to donate to
+    /// @param _amounts The amounts to donate to each recipient
+    /// @param _funder The address of the funder
```
ilpepepig, line 30: Can we remove the `_funder` param?

contracts/core/libraries/QFHelper.sol
```diff
@@ -0,0 +1,69 @@
...
+        _state.totalContributions += _totalContributionsDelta;
+    }
+
+    /// @notice Calculates the matching amount for a recipient using the Quadratic Funding formula
+    /// @param _state The state of the donations
+    /// @param _matchingAmount The total matching amount
+    /// @param _recipient The recipient to calculate the matching amount for
+    /// @return _amount The matching amount for the recipient
+    function calculateMatching(State storage _state, uint256 _matchingAmount, address _recipient)
+        internal
```
ilpepepig, line 58: Should visibility be `view` here?

contracts/core/libraries/QFHelper.sol
0xOneTony: good catch

ilpepepig (approved):

PR #615, opened 2024-07-08
feat: QV helper library

### LINKED ISSUE #608: Add QVLibrary for Quadratic Voting
## Problem
Basic mechanisms like Quadratic Voting/Funding should be abstracted into a trusted, optimized library. Provide helper functions for specific actions in different strategies.

## Solution
- Implement the logic for voting and calculating payout amount for Quadratic Voting.
- Unit Tests

## Resources

Spec for QVLibrary

```solidity
library QVLibrary {
    /// Struct that holds the state of the voting
    /// totalVoiceCredits: the total amount of voice credits casted
    /// recipientVoiceCredits: the amount of voice credits a recipient has
    /// recipientVotes: the amount of votes a recipient has
    struct VotingState {
        uint256 totalVoiceCredits;
        mapping(address => uint256) recipientVoiceCredits;
        mapping(address => uint256) recipientVotes;
    }

    /// Vote recipients
    /// Provides the recipients and amount of votes as well as the allocator state
    /// Should update the state and return it
    function voteRecipients(
		    VotingState storage _state,
        address[] memory _recipients,
        uint256[] memory _voiceCreditsToAllocate
    ) internal returns (VotingState memory) {
        // Voting logic here
        return state;
    }

    /// Calculate votes from voice credits
    function calculateVotes(uint256 _voiceCredits) internal pure returns (uint256) {
        // Calculation logic here
        return votes;
    }
    
    function calculatePayout(VotingState storage _state, address _recipient, uint256 _matchAmount) internal view returns (uint256);
}
```

0xOneTony:
Closes #608 
Quadratic Voting Library that can be used by different strategies to handle voting and calculating the payout amounts


0xAustrian (commented):
Looks good, I think some of the comments I did on https://github.com/allo-protocol/allo-v2/pull/614 apply here.

0xAustrian (approved):

contracts/core/libraries/QVHelper.sol
```diff
@@ -0,0 +1,110 @@
...
+        uint256 totalVotes;
+        uint256 totalVoiceCredits;
+        mapping(address => uint256) recipientVoiceCredits;
+        mapping(address => uint256) recipientVotes;
+    }
+
+    /// @notice Calculate the square root of a number (Babylonian method)
+    /// @param _x The number
+    /// @return _y The square root
+    function _sqrt(uint256 _x) internal pure returns (uint256 _y) {
```
thelostone-mc, line 26: Would it make sense to have this in a lib so that other strategies/helpers could use it as well ?
We also have solady which we could use https://github.com/Vectorized/solady/blob/main/src/utils/FixedPointMathLib.sol#L647

0xOneTony: agree! I will import it from a lib

0xAustrian (approved):

thelostone-mc (approved):

contracts/core/libraries/QVHelper.sol
```diff
@@ -0,0 +1,100 @@
...
+        uint256 totalVotes;
+        uint256 totalVoiceCredits;
+        mapping(address => uint256) recipientVoiceCredits;
+        mapping(address => uint256) recipientVotes;
+    }
+
+    /// @notice Votes for recipients
+    /// @param _state The voting state
+    /// @param _recipients The recipients to vote
+    /// @param _voiceCredits The amounts of voice credits to for each recipient
```
ilpepepig, line 28:
```suggestion
    /// @param _voiceCredits The amounts of voice credits to cast for each recipient
```

contracts/core/libraries/QVHelper.sol
```diff
@@ -0,0 +1,100 @@
...
+            _state.totalVoiceCredits += _voiceCredits[i];
+            /// Add the total votes
+            _state.totalVotes += _votes;
+        }
+    }
+
+    /// @notice Votes for recipients
+    /// @param _state The voting state
+    /// @param _recipients The recipients to vote
+    /// @param _votes The amounts of votes to for each recipient
```
ilpepepig, line 54:
```suggestion
    /// @param _votes The amounts of votes to cast for each recipient
```

contracts/core/libraries/QVHelper.sol
```diff
@@ -0,0 +1,100 @@
...
+        address[] memory _recipients,
+        uint256[] memory _voiceCredits
+    ) internal {
+        /// Check if the number of recipients and amounts are equal
+        if (_recipients.length != _voiceCredits.length) revert QVHelper_LengthMissmatch();
+
+        for (uint256 i = 0; i < _recipients.length; i++) {
+            /// Add the voice credits to the recipient
+            _state.recipientVoiceCredits[_recipients[i]] += _voiceCredits[i];
+            uint256 _votes = FixedPointMathLib.sqrt(_voiceCredits[i] * 1e18);
```
ilpepepig, line 41: Multiplying here by `1e18` makes `_voteWithVoiceCredits()` incompatible with `_vote()`. Is this intended?

I think we should either remove it or add it to `_vote()` as well, dividing `_voiceCredits` by `1e18`.

contracts/core/libraries/QVHelper.sol
```diff
@@ -0,0 +1,100 @@
...
+    /// @notice Calculate the payout for each recipient
+    /// @param _state The state of the Quadratic Voting
+    /// @param _recipients The recipients
+    /// @param _poolAmount The amount of the pool
+    /// @return __recipients The recipients
+    /// @return _payouts The payouts for each recipient
+    function _getPayout(VotingState storage _state, address[] memory _recipients, uint256 _poolAmount)
+        internal
+        view
+        returns (address[] memory __recipients, uint256[] memory _payouts)
```
ilpepepig, line 83: What's the reason for returning `_recipients`? Can we remove it or at least optimize it?

ilpepepig (approved):

PR #616, opened 2024-07-09
feat: implement erc2771

### LINKED ISSUE #594: Ensure core contracts support offchain voting/ meta-transactions

0xAustrian:
Closes #594

- [x] Copy erc2771 logic from ERC2771ContextUpgradeable OpenZeppelin contracts (avoided inheritance to preserve storage layout).
- [x] Add tests for `_msgData` and `_msgSender`.
- [x] Add integration test.

### DIFF
--- a/contracts/core/Allo.sol
+++ b/contracts/core/Allo.sol
@@ -79,8 +79,9 @@ contract Allo is
     /// @dev 'Pool.id' -> 'Pool'
     mapping(uint256 => Pool) private pools;
 
-    /// @custom:oz-upgrades-renamed-from cloneableStrategies
-    mapping(address => bool) private _unusedSlot;
+    /// @notice The trusted forwarder contract address
+    /// @dev Based on ERC2771ContextUpgradeable OZ contracts
+    address private _trustedForwarder;
 
     // ====================================
     // =========== Initializer =============
@@ -93,12 +94,14 @@ contract Allo is
     /// @param _treasury The address of the treasury
     /// @param _percentFee The percentage fee
     /// @param _baseFee The base fee
+    /// @param __trustedForwarder The address of the trusted forwarder
     function initialize(
         address _owner,
         address _registry,
         address payable _treasury,
         uint256 _percentFee,
-        uint256 _baseFee
+        uint256 _baseFee,
+        address __trustedForwarder
     ) external reinitializer(1) {
         // Initialize the owner using Solady ownable library
         _initializeOwner(_owner);
@@ -114,6 +117,8 @@ contract Allo is
 
         // Set the base fee
         _updateBaseFee(_baseFee);
+
+        _trustedForwarder = __trustedForwarder;
     }
 
     // ====================================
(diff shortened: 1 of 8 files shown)


0xAustrian:
Seems like integration test is failling, maybe because of old api-key in the env?

0xAustrian:
To use the erc2771 functionality a external trusted forwarder must be set. I created an example using [Biconomy](https://docs-gasless.biconomy.io/products/enable-gasless-transactions/choose-an-approach-to-enable-gasless#id-1.-standard-eip-2771-implementation-approach) Trusted Forwarder [here](https://github.com/allo-protocol/allo-v2/blob/feat/implement-erc2771/test/foundry/integration/Allo.t.sol#L55).

ilpepepig:
LGTM

I don't know what could be happening with the integration test. All other PRs are passing and they also have a test that forks from mainnet https://github.com/allo-protocol/allo-v2/blob/v2.1/test/foundry/strategies/sablier-v2/LockupBase.t.sol#L27. I'd check the api-key.

ilpepepig (approved):

contracts/core/Allo.sol
```diff
@@ -114,6 +117,8 @@ contract Allo is
 
         // Set the base fee
         _updateBaseFee(_baseFee);
+
+        _trustedForwarder = __trustedForwarder;
```
0xOneTony, line 121: I think we should be able to update this, wdyt?

0xAustrian: In the OZ implementation `_trustedForwarder` is immutable, I think they did it for some reason. But I also think we can add a function to update.

contracts/core/Allo.sol
```diff
@@ -79,8 +79,9 @@ contract Allo is
     /// @dev 'Pool.id' -> 'Pool'
     mapping(uint256 => Pool) private pools;
 
-    /// @custom:oz-upgrades-renamed-from cloneableStrategies
-    mapping(address => bool) private _unusedSlot;
+    /// @notice The trusted forwarder contract address
+    /// @dev Based on ERC2771ContextUpgradeable OZ contracts
+    address private _trustedForwarder;
```
thelostone-mc, line 84: ^ wouldn't this break upgradeability ? 
Also would it make sense to have multiple trusted forwarded ?

contracts/core/Allo.sol
```diff
@@ -653,6 +678,33 @@ contract Allo is
...
+        if (isTrustedForwarder(msg.sender) && calldataLength >= contextSuffixLength) {
+            return msg.data[:calldataLength - contextSuffixLength];
+        } else {
+            return super._msgData();
+        }
+    }
+
+    /// @dev Logic copied from ERC2771ContextUpgradeable OZ contracts
+    function _contextSuffixLength() internal view returns (uint256) {
+        return 20;
```
thelostone-mc, line 705: ^ could we not keep this constant and just use it ?

contracts/core/Allo.sol
0xAustrian: > ^ wouldn't this break upgradeability ?

It would not break it, but can introduce problems and confusions in the future. Given this I re-introduced the mapping and added a new variable.

> Also would it make sense to have multiple trusted forwarded ?

I dont think so. At least Biconomy trusted forwarder is open source and callable by anyone. There is no reason to have another one. Did you have something else in mind?

thelostone-mc: No specific reason beyond being able to use another trusted forwarder if something better comes along.
But given that we have the ability to update the forwarder , maybe it doesn't really butter

thelostone-mc (approved):

0xOneTony (approved):
LGTM! 👏

Issue #590, opened 2024-06-11
Remove receive function from Strategies


thelostone-mc:
Move receive() function over to BaseStrategy

PR #617, opened 2024-07-11
feat/core-base-strategy

### LINKED ISSUE #591: Ensure default receive is virtual and can be called only by Allo

0xOneTony:
Closes #591 , #590 

Most Basic Strategy contract that should be used for creating extensions or more complex strategies.

⚠️ It's important to note that I didn't remove the current `BaseStrategy` yet, as it would break all the strategies

### DIFF
--- a/contracts/core/interfaces/IBaseStrategy.sol
+++ b/contracts/core/interfaces/IBaseStrategy.sol
@@ -1,0 +1,119 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.19;
+
+import {IAllo} from "./IAllo.sol";
+
+interface IBaseStrategy {
+    /// ======================
+    /// ======= Errors =======
+    /// ======================
+
+    /// @notice Error when the sender is not authorized
+    error BaseStrategy_UNAUTHORIZED();
+
+    /// @notice Error when the strategy is not initialized
+    error BaseStrategy_NOT_INITIALIZED();
+
+    /// @notice Error when the strategy is already initialized
+    error BaseStrategy_ALREADY_INITIALIZED();
+
+    /// @notice Error when the pool ID is invalid
+    error BaseStrategy_INVALID_POOL_ID();
+
+    /// ======================
+    /// ======= Events =======
+    /// ======================
+
+    /// @notice Emit after recipients get registered
+    /// @param _recipient The address of the recipient
+    /// @param _data The data to use to register the recipient
+    event Registered(address indexed _recipient, bytes _data);
+
+    /// @notice Emits after recipients got allocated amount
+    /// @param _recipient The address of the recipient
+    /// @param _sender The address of the sender
+    /// @param _amount The amounts to allocate
+    /// @param _data The data to use to allocate to the recipient
+    event Allocated(address indexed _recipient, address indexed _sender, uint256 _amount, bytes _data);
+
+    /// @notice Emit after recipients got distributed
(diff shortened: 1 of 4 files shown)


contracts/core/interfaces/IBaseStrategy.sol
```diff
@@ -0,0 +1,119 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.19;
```
0xAustrian, line 2: Remember to use fixed solc version.

0xAustrian (approved):

ilpepepig (approved):

contracts/core/interfaces/IBaseStrategy.sol
```diff
@@ -0,0 +1,119 @@
...
+    /// @return _poolId The ID of the pool
+    function getPoolId() external view returns (uint256 _poolId);
+
+    /// @notice Checks the amount of tokens in the pool.
+    /// @return _poolAmount The balance of the pool
+    function getPoolAmount() external view returns (uint256 _poolAmount);
+
+    /// @notice Gets the timestamp of the phase at the given index.
+    /// @param _index The index of the phase
+    function getTimestampPhase(uint256 _index) external view returns (uint256 _timestamp);
```
thelostone-mc, line 72: What's this about ?

owocki:
you love to see it :)

what are we thinking the taxonomy of strategies should be

is it
```
basestrategy
strategy strategy strategy strategy strategy strategy 
```

or will there be multiple layers like

```
basestrategy
quadratic_straegy                              retro_strategy                        xyz_strategy
vanillaQAstrategy MACIQFstrategy            vanilla_retro                          strawberry_xyz_strategy  
```

ilpepepig:
Thinking again about this, I'm not sure about the use modifiers. I think their current logic belong in the `_beforeX()` functions and we should give custom strategy devs more flexibility.

For example, why is it assumed that `withdraw()` should be called by the Pool Manager? `_checkOnlyPoolManager` could be called from `_beforeWithdraw` if needed, and maybe by default. Also, why is the `onlyAllo` modifier not used here?

In addition, if custom strategies are not whitelisted, enforcing `onlyAllo` is not really possible but should be convinient. Can we simply remove this requirement? What would be the consequences?

Regarding `onlyInitialized`, currently strategies are cloned by Allo and initialized in the same call, which means that all strategies are by default initialized. Can we remove this modifier then? If a custom strategy requires a different initialization logic, checking that it happened could be added to `_beforeX()`.

0xOneTony:
> what are we thinking the taxonomy of strategies should be

@owocki So we should be providing 2 things to devs
1. tools to build easier their custom strategy (making use of base strategy, libraries, extensions)
2. use one of our example strategies (QVStrategy, QFStrategy, RetroPGFStrategy, RFPStrategy ....)

so the taxonomy should be 

```
BaseStrategy    /libraries   /extensions
/QuadraticVotingStrategies       /RFPStrategies      /DonationVotingStrategies
/RetroStrategies         /LTIPStrategies        /MACIStrategies
```

thelostone-mc (approved):
All look good except the reason/need for getTimestampPhase

contracts/core/interfaces/IBaseStrategy.sol
0xOneTony: yea not really needed, now that i think about it.

0xAustrian (approved):

ilpepepig (approved):

PR #618, opened 2024-07-16
feat: recipients extension

### LINKED ISSUE #619: Strategy Recipient Extension
Create an extension from `CoreBaseStrategy` that will include more detailed data for recipients along with a reviewing system.

- Should be able to define start and end time for registration, if needed
- Each recipient should have a status (none, pending, accepted, rejected, appealed)
- Should be able to register/review/update a recipient's status

0xAustrian:
Closes #619


contracts/extensions/contracts/RecipientsExtension.sol
```diff
@@ -0,0 +1,329 @@
...
+
+    /// @notice Set the recipient status.
+    /// @param _recipientId ID of the recipient
+    /// @param _status Status of the recipient
+    function _setRecipientStatus(address _recipientId, uint256 _status) internal virtual {
+        // Get the row index, column index and current row
+        (uint256 rowIndex, uint256 colIndex, uint256 currentRow) = _getStatusRowColumn(_recipientId);
+
+        // Calculate the 'newRow'
+        uint256 newRow = currentRow & ~(15 << colIndex);
```
ilpepepig, line 297: Just a personal preference: for this kind of bit operation I think using hex, 0xff instead of 15, is more readable, but both are ok.

test/utils/MockStrategyRecipientsExtension.sol
```diff
@@ -0,0 +1,472 @@
...
+        if (_success) return abi.decode(_data, (IRecipientsExtension.Status));
+        else return super._getRecipientStatus(_recipientId);
+    }
+
+    function call__getRecipientStatus(address _recipientId)
+        public
+        returns (IRecipientsExtension.Status _returnParam0)
+    {
+        return _getRecipientStatus(_recipientId);
+    }
```
ilpepepig, line 96: I'm not sure I understand these design. What is the purpose of `staticcall` and each function? Won't `success` always be false?

If the idea was to simply [expose](https://book.getfoundry.sh/tutorials/best-practices#internal-functions) the internal function I think it's enough with:
```solidity
function expose_getRecipientStatus(address _recipientId) {
   return _getRecipientStatus(_recipientId);
}
```

contracts/extensions/interfaces/IRecipientsExtension.sol
```diff
@@ -0,0 +1,70 @@
...
+        // If false, the recipientAddress is the anchor of the profile
+        bool useRegistryAnchor;
+        address recipientAddress;
+        Metadata metadata;
+    }
+
+    struct RecipientInitializeData {
+        bool metadataRequired;
+        uint64 registrationStartTime;
+        uint64 registrationEndTime;
```
thelostone-mc, line 49: ^ If we have other timings like donationStart and donationEnd time.
Those would be handled in it's own extension i assume ?

contracts/extensions/interfaces/IRecipientsExtension.sol
```diff
@@ -0,0 +1,70 @@
...
+    /// @param rowIndex The index of the row in the bitmap
+    /// @param fullRow The value of the row
+    /// @param sender The sender of the transaction
+    event RecipientStatusUpdated(uint256 indexed rowIndex, uint256 fullRow, address sender);
+
+    /// @notice Emitted when the timestamps are updated
+    /// @param registrationStartTime The start time for the registration
+    /// @param registrationEndTime The end time for the registration
+    /// @param sender The sender of the transaction
+    event TimestampsUpdated(uint64 registrationStartTime, uint64 registrationEndTime, address sender);
```
thelostone-mc, line 69: Would it be worth adding RegistrationTImestampUpdated ? I ask cause we may have other timestamps and might be nice to keep the event names distinct?  (but maybe it doesn't matter)

contracts/extensions/interfaces/IRecipientsExtension.sol
0xAustrian: That's correct

contracts/extensions/interfaces/IRecipientsExtension.sol
0xAustrian: Makes sense!

ilpepepig (approved):

0xOneTony (approved):

PR #623, opened 2024-07-19
feat: add milestones extension contract

### LINKED ISSUE #620: Strategy Milestones Extension
Create an extension from `CoreBaseStrategy` that will include defining and managing a strategy that uses milestone submissions and reviews. Strategies like `RFP` use milestones to allocate funds for recipients. We should offer an extension that includes all the logic of using milestones in strategies.

- Define milestone status
- Define milestone creation/submission/review
- Define roles for actors

ilpepepig:
Closes #620

### DIFF
--- a/contracts/extensions/contracts/MilestonesExtension.sol
+++ b/contracts/extensions/contracts/MilestonesExtension.sol
@@ -177,7 +177,9 @@ abstract contract MilestonesExtension is CoreBaseStrategy, IMilestonesExtension 
     function _validateSubmitUpcomingMilestone(address _recipientId, address _sender) internal virtual {
         // Check if the 'msg.sender' is accepted
         if (!_isAcceptedRecipient(_recipientId)) revert MilestonesExtension_INVALID_RECIPIENT();
-        if (_sender != _recipientId && !_isProfileMember(_recipientId, _sender)) revert MilestonesExtension_INVALID_SUBMITTER();
+        if (_sender != _recipientId && !_isProfileMember(_recipientId, _sender)) {
+            revert MilestonesExtension_INVALID_SUBMITTER();
+        }
 
         // Check if a submission is ongoing to prevent front-running a milestone review.
         if (milestones[upcomingMilestone].status == Status.Pending) revert MilestonesExtension_MILESTONE_PENDING();
--- a/contracts/extensions/contracts/MilestonesExtension.sol
+++ b/contracts/extensions/contracts/MilestonesExtension.sol
@@ -194,7 +194,7 @@ abstract contract MilestonesExtension is CoreBaseStrategy, IMilestonesExtension 
         if (milestones[upcomingMilestone].status != Status.Pending) revert MilestonesExtension_MILESTONE_NOT_PENDING();
     }
 
-    /// @notice Increase max bid for RFP pool
+    /// @notice Increase max bid
     /// @param _maxBid The new max bid to be set
     function _increaseMaxBid(uint256 _maxBid) internal {
         // make sure the new max bid is greater than the current max bid
--- a/contracts/extensions/contracts/MilestonesExtension.sol
+++ b/contracts/extensions/contracts/MilestonesExtension.sol
@@ -61,7 +61,7 @@ abstract contract MilestonesExtension is CoreBaseStrategy, IMilestonesExtension 
     /// ======= External/Custom =======
     /// ===============================
 
-    /// @notice Update max bid for RFP pool
+    /// @notice Update max bid
     /// @dev 'msg.sender' must be a pool manager to update the max bid.
     /// @param _maxBid The max bid to be set
     function increaseMaxBid(uint256 _maxBid) external onlyPoolManager(msg.sender) {
--- a/contracts/extensions/contracts/MilestonesExtension.sol
+++ b/contracts/extensions/contracts/MilestonesExtension.sol
@@ -5,11 +5,13 @@ pragma solidity 0.8.19;
 import {IMilestonesExtension} from "../interfaces/IMilestonesExtension.sol";
 // Core Contracts
 import {CoreBaseStrategy} from "../../strategies/CoreBaseStrategy.sol";
+// External Contracts
+import {ContextUpgradeable} from "openzeppelin-contracts-upgradeable/contracts/utils/ContextUpgradeable.sol";
 // Internal Libraries
 import {Metadata} from "../../core/libraries/Metadata.sol";
 
 /// @title Milestone Extension Strategy
-abstract contract MilestonesExtension is CoreBaseStrategy, IMilestonesExtension {
+abstract contract MilestonesExtension is CoreBaseStrategy, ContextUpgradeable, IMilestonesExtension {
     /// ================================
     /// ========== Storage =============
     /// ================================
@@ -71,7 +73,7 @@ abstract contract MilestonesExtension is CoreBaseStrategy, IMilestonesExtension 
     /// @dev Emits 'MilestonesSet' event
     /// @param _milestones Milestone[] The milestones to be set
     function setMilestones(Milestone[] memory _milestones) external virtual {
-        _validateSetMilestones(msg.sender);
+        _validateSetMilestones(_msgSender());
         uint256 totalAmountPercentage;
 
         // Loop through the milestones and add them to the milestones array
@@ -100,7 +102,7 @@ abstract contract MilestonesExtension is CoreBaseStrategy, IMilestonesExtension 
     /// @dev Emits a 'MilestonesSubmitted()' event.
     /// @param _metadata The proof of work
     function submitUpcomingMilestone(Metadata calldata _metadata) external virtual {
-        _validateSubmitUpcomingMilestone(msg.sender);
+        _validateSubmitUpcomingMilestone(_msgSender());
 
         // Get the milestone and update the metadata and status
         Milestone storage milestone = milestones[upcomingMilestone];
@@ -117,7 +119,7 @@ abstract contract MilestonesExtension is CoreBaseStrategy, IMilestonesExtension 
     /// @dev Emits a 'MilestoneStatusChanged()' event.
     /// @param _milestoneStatus New status of the milestone
     function reviewMilestone(Status _milestoneStatus) external virtual {
-        _validateReviewMilestone(msg.sender, _milestoneStatus);
+        _validateReviewMilestone(_msgSender(), _milestoneStatus);
         // Check if the milestone status is pending
 
         milestones[upcomingMilestone].status = _milestoneStatus;
@@ -155,7 +157,7 @@ abstract contract MilestonesExtension is CoreBaseStrategy, IMilestonesExtension 
     }
 
     function _validateSubmitUpcomingMilestone(address _sender) internal virtual {
-        // Check if the 'msg.sender' is accepted
+        // Check if the '_msgSender()' is accepted
         if (!_isAcceptedRecipient(_sender)) revert MilestonesExtension_INVALID_SUBMITTER();
 
         // Check if a submission is ongoing to prevent front-running a milestone review.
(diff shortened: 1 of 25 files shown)


contracts/extensions/contracts/MilestonesExtension.sol
```diff
@@ -0,0 +1,233 @@
...
+
+/// @title Milestone Strategy Extension
+abstract contract MilestonesExtension is CoreBaseStrategy, IMilestonesExtension {
+
+    /// ===============================
+    /// ========== Errors =============
+    /// ===============================
+
+    /// @notice Thrown when the milestone is invalid
+    error INVALID_MILESTONE();
```
0xOneTony, line 20: can we move the errors to the interface?

contracts/extensions/contracts/MilestonesExtension.sol
```diff
@@ -0,0 +1,233 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.19;
+
+// Interfaces
+import {IMilestonesExtension} from "../interfaces/IMilestonesExtension.sol";
+// Core Contracts
+import {CoreBaseStrategy} from "../../strategies/CoreBaseStrategy.sol";
+// Internal Libraries
+import {Metadata} from "../../core/libraries/Metadata.sol";
+import {Errors} from "../../core/libraries/Errors.sol";
```
0xOneTony, line 10: maybe we can remove this and just add a custom error which is more descriptive instead of `UNAUTHORIZED` ?

contracts/extensions/contracts/MilestonesExtension.sol
```diff
@@ -0,0 +1,189 @@
...
+    /// @param _maxBid The max bid to be set
+    function increaseMaxBid(uint256 _maxBid) external onlyPoolManager(msg.sender) {
+        _increaseMaxBid(_maxBid);
+    }
+
+    /// @notice Set the milestones for the acceptedRecipientId.
+    /// @dev Emits 'MilestonesSet' event
+    /// @param _milestones Milestone[] The milestones to be set
+    function setMilestones(Milestone[] memory _milestones) external virtual {
+        _validateSetMilestones(msg.sender);
```
0xAustrian, line 77: I think we need to use `_msgSender` here (check other places too).

contracts/extensions/contracts/MilestonesExtension.sol
```diff
@@ -0,0 +1,189 @@
...
+        if (_milestoneStatus == Status.Accepted) {
+            upcomingMilestone++;
+        }
+    }
+
+    /// ====================================
+    /// ============ Internal ==============
+    /// ====================================
+
+    function _setProposalBid(address _bidderId, uint256 _proposalBid) internal virtual {
```
0xAustrian, line 139: Can you add a small natspec on the internals too?

contracts/extensions/contracts/MilestonesExtension.sol
```diff
@@ -0,0 +1,189 @@
...
+            // If the proposal bid is 0, set it to the max bid
+            _proposalBid = maxBid;
+        }
+
+        bids[_bidderId] = _proposalBid;
+        emit SetBid(_bidderId, _proposalBid);
+    }
+
+    function _validateSetMilestones(address _sender) internal virtual {
+        _checkOnlyPoolManager(_sender);
```
0xAustrian, line 153: Why not using `onlyPoolManager` modifier? (check other places too)

contracts/extensions/contracts/MilestonesExtension.sol
```diff
@@ -0,0 +1,189 @@
...
+        // make sure the new max bid is greater than the current max bid
+        if (_maxBid < maxBid) revert MilestonesExtension_AMOUNT_TOO_LOW();
+
+        maxBid = _maxBid;
+
+        // emit the new max mid
+        emit MaxBidIncreased(maxBid);
+    }
+
+    function _getMilestonePayout(uint256 _milestoneId) internal view virtual returns (uint256) {
```
0xAustrian, line 186: This internal is never used.

0xAustrian (commented):
A few comments. I think you can use `v2.1` as base branch now. Remember to merge new changes.

contracts/extensions/contracts/MilestonesExtension.sol
ilpepepig: My rationale was that using the `onlyPoolManager` modifier was too opinionated. I guess there could be use cases in which milestones are not managed by pool managers. I think it's better to modularize the validation part of the code like this in case someone wants to customize it. Wdyt?

contracts/extensions/contracts/MilestonesExtension.sol
```diff
@@ -0,0 +1,189 @@
...
+import {Metadata} from "../../core/libraries/Metadata.sol";
+
+/// @title Milestone Extension Strategy
+abstract contract MilestonesExtension is CoreBaseStrategy, IMilestonesExtension {
+    /// ================================
+    /// ========== Storage =============
+    /// ================================
+
+    /// @notice The accepted recipient who can submit milestones.
+    address public acceptedRecipientId;
```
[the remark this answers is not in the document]
ilpepepig: My idea was that acceptedRecipientId should be set in reviewRecipients or similar, so out of scope for this extension. I'm in the middle of reviewing the RecipientsExtension though, so I'll think about this again.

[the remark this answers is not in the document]
0xAustrian: What if you remove that var and create an internal function like [this](https://github.com/allo-protocol/allo-v2/blob/a7fbf2257a8da3ce68254e11816671bc0f85c659/contracts/strategies/QVSimple.sol#L227-L232). (Make it virtual and leave the logic empty)

contracts/extensions/contracts/MilestonesExtension.sol
0xAustrian: Good point, then we can make `_checkOnlyPoolManager` (`CoreBaseStrategy.sol`) virtual so they can override and add custom logic. But that is a change for another PR.

ilpepepig: I think `_checkOnlyPoolManager` is alright as it is. I'd expect devs to modify (in this case) `_validateSetMilestones`. The one to set the milestones might be a MilestonesManager (!= PoolManager).

contracts/extensions/contracts/MilestonesExtension.sol
ilpepepig: I expect this to be used in `_distribute()`

0xAustrian (approved):

0xAustrian (approved):

thelostone-mc (approved):

contracts/extensions/contracts/MilestonesExtension.sol
```diff
@@ -0,0 +1,218 @@
...
+    mapping(address => uint256) public bids;
+
+    /// @notice Collection of milestones
+    Milestone[] internal milestones;
+
+    /// ===============================
+    /// ========= Initialize ==========
+    /// ===============================
+
+    /// @notice This initializes the BaseStrategy
```
0xOneTony, line 35:
```suggestion
    /// @notice This initializes the Milestones Extension
```

contracts/extensions/contracts/MilestonesExtension.sol
```diff
@@ -0,0 +1,218 @@
...
+    /// @dev You only need to pass the 'poolId' to initialize the BaseStrategy and the rest is specific to the strategy
```
0xOneTony, line 36:
```suggestion
    /// @dev You only need to pass the 'poolId' to initialize the Milestones Extension and the rest is specific to the strategy
```

contracts/extensions/contracts/MilestonesExtension.sol
```diff
@@ -0,0 +1,218 @@
...
+    /// @param _initializeParams The initialize params
+    function __MilestonesExtension_init(InitializeParams memory _initializeParams) internal {
+        // Set the strategy specific variables
+        _increaseMaxBid(_initializeParams.maxBid);
```
0xOneTony, line 40: since we only make use of maxBid, should we just pass as param that instead of the whole InitializeParams type?

contracts/extensions/contracts/MilestonesExtension.sol
```diff
@@ -0,0 +1,218 @@
...
+        }
+    }
+
+    /// @notice Validates if the milestone can be submitted at this moment by the `_sender`
+    /// @param _recipientId The recipient id
+    /// @param _sender The address of the submitter
+    function _validateSubmitUpcomingMilestone(address _recipientId, address _sender) internal virtual {
+        // Check if the 'msg.sender' is accepted
+        if (!_isAcceptedRecipient(_recipientId)) revert MilestonesExtension_INVALID_RECIPIENT();
+        if (_sender != _recipientId && !_isProfileMember(_recipientId, _sender)) {
```
0xOneTony, line 180: I think it is safe to remove the connection with the registry, checking if _sender is equal to _recipientId is enough and we can remove the `_isProfileMember` since it is not being used elsewhere, wdyt?

contracts/extensions/contracts/MilestonesExtension.sol
```diff
@@ -0,0 +1,218 @@
...
+    /// @notice Validates if the milestone can be reviewed at this moment by the `_sender` with `_milestoneStatus`
+    /// @param _sender The address of the reviewer
+    /// @param _milestoneStatus The new status to set the milestone to
+    function _validateReviewMilestone(address _sender, Status _milestoneStatus) internal virtual {
+        _checkOnlyPoolManager(_sender);
+        if (_milestoneStatus == Status.None) revert MilestonesExtension_INVALID_MILESTONE_STATUS();
+        if (milestones[upcomingMilestone].status != Status.Pending) revert MilestonesExtension_MILESTONE_NOT_PENDING();
+    }
+
+    /// @notice Increase max bid for RFP pool
```
0xOneTony, line 197:
```suggestion
    /// @notice Increase max bid
```

contracts/extensions/contracts/MilestonesExtension.sol
0xOneTony: since we give the option to override `_isAcceptedRecipient` the developer can add this `isProfileMember` check there

contracts/extensions/interfaces/IMilestonesExtension.sol
```diff
@@ -0,0 +1,149 @@
...
+    /// @notice Emitted when milestones are set.
+    /// @param milestonesLength Count of milestones
+    event MilestonesSet(uint256 milestonesLength);
+
+    /// ================================
+    /// =========== Enums ==============
+    /// ================================
+
+    /// @notice The Status enum that all milestones are based from
+    enum Status {
```
0xOneTony, line 69: `MilestoneStatus` I think is more clear

contracts/extensions/interfaces/IMilestonesExtension.sol
```diff
@@ -0,0 +1,149 @@
...
+
+    /// @notice Get the status of the milestone
+    /// @param _milestoneId Id of the milestone
+    function getMilestoneStatus(uint256 _milestoneId) external view returns (Status);
+
+    /// ===============================
+    /// ======= External/Custom =======
+    /// ===============================
+
+    /// @notice Update max bid for RFP pool
```
0xOneTony, line 129:
```suggestion
    /// @notice Update max bid
```

contracts/extensions/interfaces/IMilestonesExtension.sol
```diff
@@ -0,0 +1,149 @@
...
+    /// @notice Stores the details needed for initializing strategy
+    struct InitializeParams {
+        uint256 maxBid;
+    }
+
+    /// ================================
+    /// ========== Storage =============
+    /// ================================
+
+    /// @notice The maximum bid for the RFP pool.
```
0xOneTony, line 99:
```suggestion
    /// @notice The maximum bid
```

0xOneTony (commented):
🔥 Looks good! left a couple minor comments and questions

contracts/extensions/interfaces/IMilestonesExtension.sol
ilpepepig: Absolutely, I actually made that exact change in the RFPSimple branch

0xOneTony (approved):

PR #624, opened 2024-07-19
feat: gating extension

### LINKED ISSUE #622: Strategy Gating Extension
Similar to the other extensions, we want to include extra logic that can be adapted into custom strategies.

For this contract, we should support:
- gating with an erc20 and amount
- gating with an NFT
- gating with gitcoin passport
- gating with ethereum attestation service

0xOneTony:
Closes #622 DTT-252


0xAustrian (approved):
Looks good! I think the CI fails because of a linter error. Be sure to run `forge fmt`

thelostone-mc (approved):

0xKurt:
@0xOneTony 

Wouldn't it makes sense to have an extension for every type of the extension?
In this case, the abstract contracts has functions for token, nft and eas, but what happens if i want to add a new extension? For example governance token amount at a certain time point, or something similar.

Wouldn't it makes sense to split up all extensions into smaller pieces which share a common interface?

0xOneTony:
@0xKurt 

This is a fair point, it's probably cleaner if we separate the gatings into smaller extensions for each type.

I don't know about the common interface though, the whole logic is the modifier so I don't see how this can be generalized as an interface. What would the interface look like?

0xKurt:
Oh, yea you're right. As extension it maybe doesn't make sense to have a custom interface.

But in cases like gating I could see a gating contract contract for every gating type deployed which can be used across pools

E.g. token gating. One token gating contract is deployed  and the pool uses its functions to check balances, by passing token and user data.
Maybe by registering to it

0xOneTony:
@0xKurt 

Hmmm, I like the idea of the strategies registering to an external contract that can interact for gating logic. However, if we go down that path, instead of keeping them as extensions to inherit from, then splitting them into small contracts based on the gating type will give us scaling problems. We will have to keep track of MANY different gating contracts and deploy them in many different chains. In this case, maybe it's more wise to have 1 contract that supports a lot of different types.

In my opinion, I like the gating to be an extension like oz's `Ownable` but like you initially mentioned we can split them into types.

```
abstract contract TokenGatingExtension is CoreBaseStrategy
abstract contract NFTGatingExtension is CoreBaseStrategy
abstract contract EASGatingExtension is CoreBaseStrategy
```
Then users depending on their strategy they can just do

```
contract MyCustomStrategy is TokenGatingExtension, EASGatingExtensions {
   ...
   function myFunc1(..) onlyWithToken(...) { }
   function myFunc2(..) onlyWithAttestation(..) {}
   ...
}
```

This gives the option for devs to also create their own custom gating type logic and we can easily add it into our list of gating extensions in the repo. Wdyt?

cc @thelostone-mc

0xKurt:
@0xOneTony ok ok, i see your point. i totally agree with you!

contracts/extensions/EASGatingExtension.sol
```diff
@@ -0,0 +1,65 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.19;
+
+/// External Libraries
+import {IEAS, Attestation} from "eas-contracts/IEAS.sol";
+
+// Core Contracts
+import {CoreBaseStrategy} from "../strategies/CoreBaseStrategy.sol";
```
0xAustrian, line 8: You can change the path to `contracts/strategies/CoreBaseStrategy.sol`

0xAustrian (approved):

contracts/extensions/EASGatingExtension.sol
```diff
@@ -0,0 +1,65 @@
...
+    /// @notice The EAS contract
+    address public eas;
+
+    /// ===============================
+    /// ========= Initialize ==========
+    /// ===============================
+    /// @notice This initializes the strategy
+    /// @param _eas The eas address
+    function __EASGatingExtension_init(address _eas) internal {
+        /// Set the EAS contract
```
0xKurt, line 35: we should add a zero address check here

ilpepepig (approved):

0xKurt (approved):

PR #635, opened 2024-07-31
fix: reduce poolAmount in withdraw

0xAustrian:
- [x] Add a warning on `withdraw()` natspec
- [x] If the token is the `poolToken`, reduce `poolAmount`
- [x] Update tests


0xOneTony:
Closing this, since it was agreed to not reduce the pool amount. Withdraw should only be used to recover funds that might be sent by mistake

PR #638, opened 2024-08-01
test: refactor integration tests for Allo.sol and add missing tests

0xAustrian:
Closes DTT-243

- [x] Increase `reinitializer` version number
- [x] Create `IntegrationBase.sol`
- [x] Create `test_fullFlowWithMetaTx` and `test_fullFlow` test cases for `Allo.sol` tests
- [x] Add `IntegrationBase` into other integration tests
- [x] Add needed interfaces

### DIFF
--- a/test/foundry/integration/Allo.t.sol
+++ b/test/foundry/integration/Allo.t.sol
@@ -15,6 +15,8 @@ contract IntegrationAllo is IntegrationBase {
     IAllo public allo;
     DonationVotingMerkleDistributionDirectTransferStrategy public strategy;
 
+    // TODO: once other strategies are implemented, move this to the base contract
+    // Also, use IRecipientsExtension instead of the DonationVotingMerkleDistributionBaseStrategy
     function _getApplicationStatus(address _recipientId, uint256 _status, address payable _strategy)
         internal
         view
--- a/test/foundry/integration/Allo.t.sol
+++ b/test/foundry/integration/Allo.t.sol
@@ -25,7 +25,14 @@ contract IntegrationAllo is Test {
     address public treasury;
     address public userAddr;
 
+    address public recipient0Addr;
+    address public recipient1Addr;
+    address public recipient2Addr;
+
     uint256 public userPk;
+    uint256 public recipient0Pk;
+    uint256 public recipient1Pk;
+    uint256 public recipient2Pk;
 
     bytes32 public profileId;
 
@@ -40,7 +47,11 @@ contract IntegrationAllo is Test {
 
         owner = makeAddr("owner");
         treasury = makeAddr("treasury");
+
         (userAddr, userPk) = makeAddrAndKey("user");
+        (recipient0Addr, recipient0Pk) = makeAddrAndKey("recipient0");
+        (recipient1Addr, recipient1Pk) = makeAddrAndKey("recipient1");
+        (recipient2Addr, recipient2Pk) = makeAddrAndKey("recipient2");
 
         allo = new Allo();
         registry = new Registry();
@@ -57,6 +68,23 @@ contract IntegrationAllo is Test {
 
         // Deal 100k DAI to the user
         deal(dai, userAddr, 100_000 ether);
+    }
+
+    function _getApplicationStatus(address _recipientId, uint256 _status, address payable _strategy)
+        internal
+        view
+        returns (DonationVotingMerkleDistributionDirectTransferStrategy.ApplicationStatus memory)
+    {
+        uint256 recipientIndex =
+            DonationVotingMerkleDistributionBaseStrategy(_strategy).recipientToStatusIndexes(_recipientId) - 1;
+
+        uint256 rowIndex = recipientIndex / 64;
+        uint256 colIndex = (recipientIndex % 64) * 4;
+        uint256 currentRow = DonationVotingMerkleDistributionBaseStrategy(_strategy).statusesBitMap(rowIndex);
+        uint256 newRow = currentRow & ~(15 << colIndex);
+        uint256 statusRow = newRow | (_status << colIndex);
+
+        return DonationVotingMerkleDistributionBaseStrategy.ApplicationStatus({index: rowIndex, statusRow: statusRow});
     }
 
     /// @dev Send a transaction using the Biconomy Forwarder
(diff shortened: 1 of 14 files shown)


0xOneTony (approved):
LGTM! 👏 would love to see some more comments on each step of the full flows, maybe also explaining some of the numbers you assert in the end aswell. One thing to note is when we update Allo to use the signatures of the new `CoreBaseStrategy` the tests might need a slight refactor and it would be great to use a strategy from the refactored ones but we will leave this for an other PR.

0xOneTony:
Also let's fix the workflow, seems there is a format issue @0xAustrian

ilpepepig:
Should we add tests for the upgrade here? If I understand correctly `initialize()` https://github.com/allo-protocol/allo-v2/blob/4628232f9105db806180d8833775411f9a0c992e/test/foundry/integration/Allo.t.sol#L62 will actually revert if the upgrade is attempted with the current implementation because of this: https://github.com/allo-protocol/allo-v2/blob/4628232f9105db806180d8833775411f9a0c992e/contracts/core/Allo.sol#L108 The new version should be 2 instead 1.

0xAustrian:
> Should we add tests for the upgrade here? If I understand correctly `initialize()`
> 
> https://github.com/allo-protocol/allo-v2/blob/4628232f9105db806180d8833775411f9a0c992e/test/foundry/integration/Allo.t.sol#L62
> 
> will actually revert if the upgrade is attempted with the current implementation because of this:
> https://github.com/allo-protocol/allo-v2/blob/4628232f9105db806180d8833775411f9a0c992e/contracts/core/Allo.sol#L108
> 
> The new version should be 2 instead 1.

Interesting, yes. Would be cool to deploy as an update.

test/foundry/integration/ITransparentUpgradeableProxy.sol
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.19;
+
+interface ITransparentUpgradeableProxy {
```
0xOneTony, line 4: Cant we import these interfaces instead of adding them into the repo?

test/foundry/integration/IntegrationBase.sol
```diff
@@ -0,0 +1,48 @@
...
+
+    uint256 public percentFee;
+    uint256 public baseFee;
+
+    function setUp() public virtual {
+        vm.createSelectFork(vm.rpcUrl("mainnet"), 20289932);
+
+        // Get the current protocol variables
+        // TODO: call admin() instead of reading storage 
+        // alloAdmin = ITransparentUpgradeableProxy(ALLO_PROXY).admin();
```
0xOneTony, line 26: this doesnt work?

test/foundry/integration/Allo.t.sol
```diff
@@ -1,120 +1,512 @@
...
     DonationVotingMerkleDistributionDirectTransferStrategy public strategy;
 
-    address public owner;
     address public relayer;
-    address public treasury;
     address public userAddr;
 
+    address public recipient0Addr;
+    address public recipient1Addr;
+    address public recipient2Addr;
```
0xOneTony, line 27: We should move recipients to the IntegrationBase aswell and the token we use in this case dai IMO

test/foundry/integration/Allo.t.sol
```diff
@@ -1,120 +1,512 @@
...
             address(allo), "Test Strategy", ISignatureTransfer(address(1))
         );
 
-        allo.initialize(owner, address(registry), payable(treasury), 0, 0, biconomyForwarder);
-        registry.initialize(owner);
-
         vm.prank(userAddr);
-        profileId =
-            registry.createProfile(0, "Test Profile", Metadata({protocol: 1, pointer: ""}), userAddr, new address[](0));
+        profileId = IRegistry(registry).createProfile(
```
0xOneTony, line 56: let's also handle profile creations in the integration base

test/foundry/integration/ITransparentUpgradeableProxy.sol
0xAustrian: This one is not available in OZ package

test/foundry/integration/IntegrationBase.sol
0xAustrian: Nope, for some reason it fails. I will try debug it again.

ilpepepig: AFAIK `admin()` is only callable by the admin itself. So, if you are the admin you can call `admin()` to see that you are the admin. If you are not the admin you won't be able to know who the admin is, but you can check that you are not the admin. Super intuitive.

The admin is usually a ProxyAdmin contract which address should be known and which can be an admin of several proxies. In this case the ProxyAdmin is [0x758b87af7fdB4783f848a1dDEa1F025dC48B9858](https://etherscan.io/address/0x758b87af7fdB4783f848a1dDEa1F025dC48B9858) and has `0x34d82D1ED8B4fB6e6A569d6D086A39f9f734107E` as owner.

ilpepepig: If the ProxyAdmin address is not known, it can usually be found by looking at the contract creation transaction logs. I don't know if there is a better way.

ilpepepig (approved):

thelostone-mc (approved):

test/foundry/integration/IntegrationBase.sol
thelostone-mc: I know on hardhat we could regenerate the our zeppelin file , is there no way to do that for foundry ?

test/foundry/integration/Allo.t.sol
```diff
@@ -1,58 +1,62 @@
...
+    IAllo public allo;
     DonationVotingMerkleDistributionDirectTransferStrategy public strategy;
 
-    address public owner;
-    address public relayer;
-    address public treasury;
-    address public userAddr;
-
-    uint256 public userPk;
+    // TODO: use a new strategy and remove this function
```
0xAustrian, line 18: To solve this we have to wait for #641 to be merged.

test/foundry/integration/IntegrationBase.sol
0xAustrian: Sorry!, missed this comment. 
I'm not sure if that is possible. Anyway, we can query the admin reading storage so is not a big deal.

0xOneTony (approved):
LGTM!

thelostone-mc (approved):

PR #639, opened 2024-08-01
feat: make reviewRecipients more customizable

ilpepepig:
The idea of these changes is that:
- It should be easy to add custom review logic to each new Status being set in `reviewRecipients()` (e.g. do sth on Status.Reject).
- It should be easy to differentiate a new Status from an old Status. Currently in `reviewRecipients()` it is not straight forward to see which statuses are new and which are not, because a row with 64 statuses is passed as param and set to storage directly.
- It should be possible and easy to access the recipient id of the new Status being set.
- It should be easy to override `reviewRecipients()` to use a different permission logic.
- We should keep the gas optimization of using a bitmap for recipients statuses.

Some ideas implemented so far:
- [x] merge and pack recipientToStatusIndexes into Recipient struct
- [x] add mapping containing the `recipientIndex` --> `recipientId` information.
- [x] iterate all the statuses in reviewRecipient and call an internal customizable function. Only if an immutable var was set to true to save gas when this is not needed.

### DIFF
--- a/contracts/extensions/contracts/RecipientsExtension.sol
+++ b/contracts/extensions/contracts/RecipientsExtension.sol
@@ -50,7 +50,8 @@ abstract contract RecipientsExtension is CoreBaseStrategy, Errors, IRecipientsEx
 
     /// @notice Constructor to set the Allo contract
     /// @param _allo Address of the Allo contract.
-    constructor(bool _reviewEachStatus) {
+    /// @param _reviewEachStatus true if custom review logic was added.
+    constructor(address _allo, bool _reviewEachStatus) CoreBaseStrategy(_allo) {
         REVIEW_EACH_STATUS = _reviewEachStatus;
     }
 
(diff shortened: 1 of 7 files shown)


contracts/extensions/contracts/RecipientsExtension.sol
```diff
@@ -101,29 +113,45 @@ abstract contract RecipientsExtension is CoreBaseStrategy, Errors, IRecipientsEx
...
+                    uint8 currentStatus = uint8((currentRow >> col * 4) & 0xF);
+
+                    if (newStatus != currentStatus) {
+                        uint256 recipientIndex = rowIndex * 64 + col + 1;
+                        _reviewRecipientStatus(Status(newStatus), recipientIndex);
+                    }
+
+                    unchecked {
+                        col++;
+                    }
```
0xAustrian, line 138: `col` is incremented here and in the `line 127` is that correct?

contracts/strategies/RFPSimple.sol
```diff
@@ -61,44 +61,23 @@ contract RFPSimple is CoreBaseStrategy, MilestonesExtension, RecipientsExtension
...
 
+    function _validateReviewRecipients(address _sender) internal virtual override {
+        _checkOnlyActiveRegistration();
+        super._validateReviewRecipients(_sender);
+    }
+
+    function _reviewRecipientStatus(Status _newStatus, uint256) internal virtual override {
+        if (_newStatus == IRecipientsExtension.Status.Accepted) {
+            if (block.timestamp > registrationEndTime) revert INVALID();
+            // THe registration pariod ends when a recipient is accepted.
```
0xAustrian, line 76: _THe registration pariod_ -> _The registration period_

contracts/extensions/contracts/RecipientsExtension.sol
```diff
@@ -101,29 +113,45 @@ abstract contract RecipientsExtension is CoreBaseStrategy, Errors, IRecipientsEx
...
+                }
+            }
+
```
0xAustrian, line 141: Can we add a test for this new functionality?

0xAustrian (approved):

0xOneTony (approved):

thelostone-mc (approved):

PR #641, opened 2024-08-06
feat: update Allo to comply new CoreBaseStrategy.sol

0xAustrian:
Closes DTT-259

- [x] Update `Allo.sol`
- [x] Fix tests in `Allo.t.sol`
- [x] Temporarily comment out old strategy tests

### DIFF
--- a/contracts/core/Allo.sol
+++ b/contracts/core/Allo.sol
@@ -178,7 +178,7 @@ contract Allo is
             _msgSender(),
             msg.value,
             _profileId,
-            IStrategy(_strategy),
+            IBaseStrategy(_strategy),
             _initStrategyData,
             _token,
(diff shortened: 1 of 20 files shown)


ilpepepig:
Not sure if relevant for this PR, but I think we should use `calldata` almost everywhere or not use it at all for consistency. I'd rather use `calldata` to save some gas.

Otherwise LGTM!

thelostone-mc (approved):

thelostone-mc:
> Not sure if relevant for this PR, but I think we should use calldata almost everywhere or not use it at all for consistency. I'd rather use calldata to save some gas.

Do it in a follow up PR ?

ilpepepig (approved):

PR #643, opened 2024-08-09
test: add .tree for Allo.sol

0xAustrian:
Closes DTT-281


contracts/core/Allo.sol
```diff
@@ -716,7 +716,7 @@ contract Allo is
     }
 
     /// @dev Logic copied from ERC2771ContextUpgradeable OZ contracts
-    function _msgData() internal view virtual override returns (bytes calldata) {
```
0xAustrian, line 719: I had to remove the `virtual` keyword so smock does not crash when generating the mock. Looks like the return paramater of type `calldata` is messing with the generation.

test/foundry/core/AlloUnit.tree
```diff
@@ -0,0 +1,287 @@
...
+    └── it should call removePoolManagers
+
+Allo::recoverFunds
+├── when sender is not owner
+│   └── it should revert
+└── when sender is owner
+    ├── when token is native
+    │   └── it should transfer native token
+    └── when token is not native
+        └── it should transfer token
```
0xOneTony, line 85: for the tests let's make sure we also check it transfers the correct amount

package.json
```diff
@@ -29,7 +29,8 @@
     "deploy-deployer": "npx hardhat run scripts/core/deployContractFactory.ts --network",
     "/////// deploy-test ///////": "echo 'deploy test scripts'",
     "create-profile": "npx hardhat run scripts/test/createProfile.ts --network",
-    "create-pool": "npx hardhat run scripts/test/createPool.ts --network"
+    "create-pool": "npx hardhat run scripts/test/createPool.ts --network",
+    "smock": "smock-foundry --contracts contracts/core"
```
0xOneTony, line 33: Should we specify the full path? we currently only want Allo.sol to get generated

contracts/core/Allo.sol
```diff
@@ -464,14 +464,14 @@ contract Allo is
     /// @notice Internal function to check is caller is pool manager
     /// @param _poolId The pool id
     /// @param _address The address to check
-    function _checkOnlyPoolManager(uint256 _poolId, address _address) internal view {
+    function _checkOnlyPoolManager(uint256 _poolId, address _address) internal view virtual {
```
0xOneTony, line 467: Why do we add virtual in all these calls? Is it needed for smock?

package.json
0xAustrian: I tried but fails with `[Error: ENOTDIR: not a directory, scandir '/.../.../.../.../contracts/core/Allo.sol']`. I think smock only allows folders

contracts/core/Allo.sol
0xAustrian: [Yes!](https://github.com/defi-wonderland/smock-foundry?tab=readme-ov-file#gotchas)

test/foundry/core/AlloUnit.tree
0xAustrian: Yes, here I'm assuming that we will add a `expectCall` with exact parameters.

thelostone-mc (approved):

0xOneTony (approved):

PR #649, opened 2024-08-21
feat: make Transfer.sol a library and improve usage

ilpepepig:


contracts/strategies/DirectAllocation.sol
```diff
@@ -69,15 +74,23 @@ contract DirectAllocationStrategy is CoreBaseStrategy {
...
+        uint256 _totalNativeAmount;
         for (uint256 _i = 0; _i < _recipientsLength;) {
             /// Direct allocate the funds
-            _transferAmountFrom(_tokens[_i], TransferData({from: _sender, to: _recipients[_i], amount: _amounts[_i]}));
+            if (_tokens[_i] == NATIVE) {
+                _totalNativeAmount += _amounts[_i];
+                _recipients[_i].safeTransferETH(_amounts[_i]);
+            } else {
+                _tokens[_i].safeTransferFrom(_sender, _recipients[_i], _amounts[_i]);
+            }
```
0xOneTony, line 85: Why not use the `transferAmount` here and only have an if for accounting the `_totalNativeAmount` ?

contracts/strategies/DonationVotingOffchain.sol
```diff
@@ -275,13 +280,13 @@ contract DonationVotingOffchain is CoreBaseStrategy, RecipientsExtension {
             if (tokens[i] == NATIVE) {
                 totalNativeAmount += _amounts[i];
             } else {
-                SafeTransferLib.safeTransferFrom(tokens[i], _sender, address(this), _amounts[i]);
+                tokens[i].safeTransferFrom(_sender, address(this), _amounts[i]);
```
0xOneTony, line 283: Same q here, and we can remove the `SafeTransferLib`

contracts/strategies/DirectAllocation.sol
ilpepepig: `transferAmount` uses `safeTransfer` not `safeTransferFrom`. We could add a function like the following:

```solidity
    function transferAmountFrom(address _token, address _from, address _to, uint256 _amount) internal {
        if (_token == NATIVE) {
            // '_from' is ignored. The contract's balance is used. 
            if (_to != address(this)) _to.safeTransferETH(_amount);
        } else {
            _token.safeTransferFrom(_from, _to, _amount);
        }
    }
```

It would be slightly inefficient and the dev won't be completely abstracted away from what's happening (native balance still needs to be checked), but `SafeTransferLib` won't be needed and the code might look simpler in some cases.

I was reluctant to keep transferAmountFrom, but thinking again about it I like it. Wdyt?

0xOneTony: I think based on the use cases we currently have it makes sense to have the `transferAmountFrom`, probably worth adding it back

0xAustrian (approved):

thelostone-mc (approved):

PR #651, opened 2024-08-25
feat: upgrade solidity 0.8.22

ilpepepig:
Closes DTT-289

About the decision to use ^0.8.19 instead of locking 0.8.22.

### The good: don't force integrators to use 0.8.22
1. Extensions and strategies are meant to be used and inherited by integrators. They might need to use other external contracts as well that have different compiler requirements, for example hedgey vesting contracts or a particular openzeppelin release. For this, locking the compiler version could be limitting.
2. Integrators can use the latest compiler versions, reducing vulnerability risks and benefitting from optimizations and new features (transient storage for example).
3. Most EVM chains support `PUSH0` by now, which was introduced in 0.8.20 (Shanghai). However, in case someone wants to deploy on a chain which doesn't, allowing to use solidity 0.8.19 gives flexibility to easily do so.

### The bad
1. Even non-breaking updates to solidity can introduce subtle changes that potentially lead to vulnerabilities, so auditing contracts that support many compiler versions requires more effort and won't be able to predict the contracts behavior under new upgrades.


0xAustrian (approved):
Looks good!

0xOneTony (approved):

thelostone-mc (approved):
LGTM but curious if we it make sense to bump the version for others `pragma solidity ^0.8.19;` to `pragma solidity ^0.8.22;` 
Would there be side-effects cause of this ? @ilpepepig

ilpepepig:
> LGTM but curious if we it make sense to bump the version for others `pragma solidity ^0.8.19;` to `pragma solidity ^0.8.22;` Would there be side-effects cause of this ? @ilpepepig

My main concern was `PUSH0` not being supported everywhere, but looking into this https://push0.info/ it seems like all the relevant chains support it already. I guess it won't be an issue to use `pragma solidity ^0.8.22` after all.

Other than that, the more compiler versions we support, the more likely integrators won't have issues integrating extensions and strategies with other third party contracts that might lock their contracts at a lower solidity version.

I think I'd rather leave it at `pragma solidity ^0.8.19;` but I don't have a strong opinion on this. Whatever the rest thinks it's best I'm ok.

0xOneTony:
I think it's okey to keep it `^0.8.19` like ilpepe mentioned, we have defined 0.8.22 on foundry.toml so I think this is enough

PR #652, opened 2024-08-26
refactor: repo structure

0xOneTony:
The restructure of the repo is described here: https://www.notion.so/defi-wonderland/Repository-Restructure-b63179e113884c79b6f797f5db940109

Also updated remappings and imports.

Decided to delete a lot of test files and test helper contracts and mocks from strategies that are in the `deprecated` folder. Things get complex if we try to have tests for old and new strategies. `v2.1` should only include tests for core and the refactored strategies, extensions and libraries.


contracts/core/Registry.sol
```diff
@@ -2,17 +2,18 @@
...
 // External Libraries
-import "openzeppelin-contracts-upgradeable/contracts/access/AccessControlUpgradeable.sol";
-import "openzeppelin-contracts-upgradeable/contracts/proxy/utils/Initializable.sol";
+import {AccessControlUpgradeable} from
+    "openzeppelin-contracts-upgradeable/contracts/access/AccessControlUpgradeable.sol";
+import {Initializable} from "openzeppelin-contracts-upgradeable/contracts/proxy/utils/Initializable.sol";
 import {ERC20} from "solady/tokens/ERC20.sol";
 // Interfaces
-import "./interfaces/IRegistry.sol";
+import {IRegistry} from "./interfaces/IRegistry.sol";
```
0xAustrian, line 10: (This can be in another PR) Would be cool to remove relative paths and just use the remappings.

test/utils/Time.sol
0xAustrian, line 1: I think this file is not in use.

contracts/core/Registry.sol
0xOneTony: 100% I started changing but it was popping up more issues on certain contracts so I left for an other PR the rest. I will open a ticket

0xAustrian (approved):

foundry.toml
```diff
@@ -6,9 +6,8 @@ optimizer-runs = 1_000
 src = 'contracts'
 out = 'out'
 libs = ['node_modules', 'lib']
-test = 'test/foundry'
+test = 'test'
 cache_path = 'cache_forge'
-remappings = []
 
 # Hedgey Profile
```
ilpepepig, line 12: The hedgey strategies tests required this, but it's no longer needed. It can be removed along with its usage here https://github.com/allo-protocol/allo-v2/blob/3ce91e28644a2f29109f2416bb5a5c8a89eed096/package.json#L11-L14

0xAustrian (approved):

ilpepepig (approved):

thelostone-mc (approved):

PR #654, opened 2024-08-26
feat: add AllocationExtension.sol

0xAustrian:


test/utils/mocks/MockAllocationExtension.sol
0xAustrian, line 1: I had to create another mocks folder with a custom mock for the extension, since Smock does not work well with abstract contracts.

contracts/extensions/contracts/AllocationExtension.sol
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    /// ====================================
+    /// ============ Internal ==============
+    /// ====================================
+
+    /// @notice Checks if the allocator is valid
+    /// @param _allocator The allocator address
+    /// @return 'true' if the allocator is valid, otherwise 'false'
+    function _isValidAllocator(address _allocator) internal view virtual returns (bool);
```
0xOneTony, line 84: I think we can remove this, it should be part of the allowlist extension

test/foundry/extensions/AllocationExtension.t.sol
0xOneTony, line 1: Let's follow the example of the AlloUnit.t.sol with the naming, so it is easier as well to filter with forge only the unit tests. So we should change the file name to `AllocationExtensionUnit.t.sol`

0xOneTony (commented):
Couple small comments, other than that looks good!

contracts/extensions/contracts/AllocationExtension.sol
0xAustrian: My idea is that maybe you need `_isValidAllocator` but dont need all other logic from allowlist extension. 
For example in: `DirectGrantsSimpleStrategy`, `MicroGrantsGovStrategy`, `MicroGrantsHatsStrategy`, and a few other strats

0xOneTony: oke sounds good then

ilpepepig: What about implementing it and returning `true` as default so that strategies not using `_isValidAllocator()` don't have to implement it.

contracts/extensions/contracts/AllocationExtension.sol
```diff
@@ -0,0 +1,141 @@
...
+
+        allocationStartTime = _allocationStartTime;
+        allocationEndTime = _allocationEndTime;
+
+        emit AllocationTimestampsUpdated(allocationStartTime, allocationEndTime, msg.sender);
+    }
+
+    /// @dev Ensure the function is called before allocation start time
+    function _checkBeforeAllocation() internal virtual {
+        if (block.timestamp > allocationStartTime) revert ALLOCATION_HAS_STARTED();
```
ilpepepig, line 112: Allocation can currently be active and "before allocation" at the same time.

```suggestion
        if (block.timestamp >= allocationStartTime) revert ALLOCATION_HAS_STARTED();
```

ilpepepig (approved):

contracts/extensions/contracts/AllocationExtension.sol
```diff
@@ -0,0 +1,141 @@
...
+    /// @dev The 'msg.sender' must be a pool manager.
+    /// @param _allocationStartTime The start time for the allocation
+    /// @param _allocationEndTime The end time for the allocation
+    function _updateAllocationTimestamps(uint64 _allocationStartTime, uint64 _allocationEndTime) internal virtual {
+        if (_allocationStartTime > _allocationEndTime) revert INVALID_ALLOCATION_TIMESTAMPS();
+
+        allocationStartTime = _allocationStartTime;
+        allocationEndTime = _allocationEndTime;
+
+        emit AllocationTimestampsUpdated(allocationStartTime, allocationEndTime, msg.sender);
```
ilpepepig, line 107: To avoid reading from storage:
```suggestion
        emit AllocationTimestampsUpdated(_allocationStartTime, _allocationEndTime, msg.sender);
```

thelostone-mc (approved):
lgtm. added a nitpick but not important

contracts/extensions/contracts/AllocationExtension.sol
```diff
@@ -0,0 +1,141 @@
...
+    /// @return 'true' if the allocator is valid, otherwise 'false'
+    function _isValidAllocator(address _allocator) internal view virtual returns (bool);
+
+    /// @notice Returns TRUE if the token is allowed
+    /// @param _token The token to check
+    function _isAllowedToken(address _token) internal view virtual returns (bool) {
+        // all tokens allowed
+        if (allowedTokens[address(0)]) return true;
+
+        if (allowedTokens[_token]) return true;
```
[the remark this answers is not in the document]
ilpepepig: Splitting the `if` in two instead of using the or `||` saves a bit of gas if I remember correctly.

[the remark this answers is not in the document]
0xAustrian: Because of gas saving and readability I will keep it as is.

0xOneTony (approved):

PR #658, opened 2024-09-04
feat: add strategies deployment script

0xAustrian:
Closes DTT-296

PR #660, opened 2024-09-09
feat: allo v2.1 deployment script

ilpepepig:


script/core/DeployAllo.sol
0xAustrian, line 1: In #658 I created `DeployBase.sol`, do you think you can make use of it?
Also, @0xOneTony suggested to set the deploy vars on the `.env` so it's easier to change and deploy, and I think here makes even more sense since the variables barely change within chains. wdyt?

ilpepepig: - [x] Inheriting from `DeployBase.sol` https://github.com/allo-protocol/allo-v2.1/pull/3/commits/5f983c115dffe340dad4fd15f4c2cff67812ad7f

I'm not sure about moving vars to `.env`. Why would that be easier? Deploying on different chains would require you change the `.env` deployment vars instead of simply calling `bun run deploy-allo-v2.1 <CHAIN>`, right? It would also make it more difficult to deploy to N chains in loop if that was necessary.

